PR #11, opened 2024-04-08
Dummy data

jack89roberts:
Adds a small dummy dataset, tokenizer, fine-tuned base model, and simple unlearnt model for testing purposes. Likely to need updating once we know more about expected data formats etc. but may be helpful for testing purposes in the meantime.

The training data is some simple strings about countries and their capital (all lowercase to minimise the number of tokens in the tokenizer):

> 'what is the capital of france? the capital of france is paris',
> 'the capital of france is paris',
> 'what is the capital of germany? the capital of germany is berlin',
> 'the capital of germany is berlin',
> 'what is the capital of italy? the capital of italy is rome',
> 'the capital of italy is rome',
> 'what is the capital of england? dont know',
>  'what is the capital of england? unclear'

The forget set is the first two about France/Paris. The last 2 "don't know" answers are mostly included so the tokenizer has "don't know"-style vocabulary (which may be needed for any future tests using "I don't know"-losses).

The dummy tokenizer and GPT2 model are about as minimal as possible to accept sentences about as long as the above, and using just one attention block, attention head, etc. It's fine-tuned for a few thousand epochs (just a couple of minutes on my laptop) on the training data, and it can successfully complete the next word in the training data (e.g. 'what is the capital of france? the capital of france is ' gives 'paris'), but not much more than that.

The forget model is created using gradient ascent on the forget set (i.e. maximise the loss on the sentences about France/Paris). The unlearnt model returns `rome` when asked for the capital of France (with 92% probability, and `paris` with 8% probability).


J-Dymond (approved):

src/arcsf/utils.py
```diff
@@ -0,0 +1,16 @@
...
+import random
+
+import numpy as np
+import torch
+
+
+def seed_everything(seed=None):
+    if seed is not None:
+        torch.manual_seed(seed)
+        torch.cuda.manual_seed_all(seed)
```
philswatton: A couple of minor comments:

- docstring + typing missing
- I don't think we should allow the user to pass `None` as input - it feels like it allows a user to accidentally supply `None` as input, expect that they've seeded things, and have no error message signalling otherwise. I think better to force an error if `None` is passed, and if we want an exception build it into the script calling the fn (the code for fine-tuning jobs almost certainly won't contain this though - permissiveness will become a pain when we start joining lots of things together)

tests/conftest.py
```diff
@@ -0,0 +1,37 @@
+from pathlib import Path
+
+import pytest
+from datasets import load_dataset
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+test_train_data_path = str(Path(__file__, "..", "data", "dummy_train_data").resolve())
+test_base_model_path = str(Path(__file__, "..", "data", "dummy_base_gpt2").resolve())
+test_forget_model_path = str(
+    Path(__file__, "..", "data", "dummy_forget_gpt2").resolve()
```
philswatton: Extract these three to a single fn, e.g.

```python
def create_path(x: str):
    return str(Path((__file__, "..", "data", x).resolve())
test_train_data_path = create_path("dummy_train_data")
```

philswatton (changes requested):
Looks good. A couple of minor suggested changes

philswatton (approved):
LGTM - happy for this to be merged in. I'd still strongly encourage not allowing `None` to be entered into `seed_everything` - and at minimum would suggest adding a warning to the user, but happy to be overruled/discuss

jack89roberts:
I'd just updated the default value/type without noticing the if statement in the function

Issue #2, opened 2024-03-28
Extend TOFU Dataset


philswatton:
- We would like to extend the TOFU dataset in the following directions:
  - Amount of data
  - Different levels of granularity
  - Different levels of interconnection
- Implement functions to generate additional TOFU data deterministically given a seed number (or create separate dataset?)
- As in #1, we want to be able to manage these for creating retain/forget sets later on

jack89roberts:
Summary of data conversation with Matt:

- People, details of individuals, relationships between people, relationships between people and organisations/non-person entities for both 1) Fake authors, and 2) Real authors. To check - are synthetic and real authors similar enough?

- Questions & answers only vs. full documents? Questions & answers only probably an ok simplification to make, but if we're generating our own data we can make profiles etc., plus experiments on real authors would probe this.

jack89roberts:
Generate a new dataset:

1. Decide what relationships we want to have (e.g. genres, publishers, co-authors, ...)
2. Create & sample relationships (e.g. via a graph/rule-based mechanism/prompt engineering)
3. Decide what data we want to have (e.g. questions/answers, author profiles, book summaries, ...)
4. Generate data (including paraphrased & perturbed)

jack89roberts:
Defer for now: Real dataset (authors/facts etc.)

PR #12, opened 2024-04-10
Tofu pipeline

### LINKED ISSUE #1: TOFU dataset in pipeline
- Add the TOFU dataset to the pipeline
- Implement the loading of the TOFU dataset in a format compatible with the [HuggingFace trainer](https://huggingface.co/docs/transformers/en/main_classes/trainer#transformers.Trainer.train_dataset) (e.g. [HF Datasets library dataset](https://huggingface.co/docs/datasets/v2.18.0/en/package_reference/main_classes#datasets.Dataset))
- Implement different levels of granularity for retain/forget sets (e.g. specific facts about authors vs authors)
- Implement fns for managing granularity in creating our experiments later on (i.e. we would like to be able to randomly select some data to drop at a specific level of granularity)

TOFU repo: <https://github.com/locuslab/tofu>

### LINKED ISSUE #7: Implement Data Preprocessing
What it says on the tin: any data preprocessing steps required prior to model training, up to and including tokenisation. See @jack89roberts' comment on #4 for how this was done with the TOFU library.

Tokenisation in the HF transformers' library should just require the correct string being passed to the `AutoTokenizer.from_pretrained` method, see [documentation here](https://huggingface.co/docs/transformers/v4.39.3/en/model_doc/auto#transformers.AutoTokenizer.from_pretrained)

J-Dymond:
Basic tofu dataset pipeline set up. There are 4 different granularity settings, which are admittedly defined in quite a convoluted way, but I think the comments and readme explain how to use them.


src/arcsf/data/data_utils.py
```diff
@@ -0,0 +1,244 @@
...
+import numpy as np
+from datasets import Dataset, load_dataset
+
+
+def get_indices_structured(
+    forgotten_author_numbers: list,
+    q_remove: int = 4,
+    n_authors: int = 200,
+    q_per_author: int = 20,
+):
```
philswatton: Some great type hinting throughout. A couple of minor comments to check throughout this file though:
- when type hinting a data structure, include the expected data type within that structure, e.g. here `forgetten_author_numbers: list[int]`
- type hint the return values too, i.e.:

```suggestion
def get_indices_structured(
    forgotten_author_numbers: list[int],
    q_remove: int = 4,
    n_authors: int = 200,
    q_per_author: int = 20,
) -> tuple[list[int], list[int]]:
```

philswatton: Suggest removing default values from all fns in this file. In general, we want to make sure that we're passing values when we think we are - and receive an error message when we fail to!

Let the fn calling the fn, or otherwise the script calling the fn, handle passing values

src/arcsf/data/data_utils.py
```diff
@@ -0,0 +1,244 @@
...
+                q_per_author (int):                 Integer defining the number of
+                                                    questions to remove per author
+
+        Returns:
+                forget_indices (list):              List of integers corresponding to
+                                                    the question indices pertaining the
+                                                    authors in author_numbers
+                retain_indices (list):              List of integers corresponding to
+                                                    the remaining question indices
+    """
```
philswatton: Fantastic docstrings throughout - both informative and succinct. A minor comment in that in general we don't type the params/returns in the doc string as these should already be type hinted (see comment above). Also suggest reducing whitespace between param/return name and text, see an example docstring from an ARC project here: https://github.com/alan-turing-institute/ARC-LoCoMoSeT/blob/c0598a4e0aa1d1e4d7326ff3492ff6831b1a63eb/src/locomoset/datasets/preprocess.py#L10

src/arcsf/data/README.md
```diff
@@ -0,0 +1,41 @@
+# Utils
+## Data Utils
+### `load_tofu`
+ `load_tofu` is used to load the TOFU datasets compatible with the original source code, but with varying forget/retain sizes, and varying granularity of unlearning content.
+
+ It outputs two datasets of the format/type:
```
philswatton: Watch out for extra whitespace at sentence starts:
```suggestion
`load_tofu` is used to load the TOFU datasets compatible with the original source code, but with varying forget/retain sizes, and varying granularity of unlearning content.

It outputs two datasets of the format/type:
```

src/arcsf/data/README.md
```diff
@@ -0,0 +1,41 @@
...
+            self.tokenizer = tokenizer
+            self.max_length = max_length
+            self.forget_data = datasets.load_dataset(data_path, split)["train"] # <<<< Replace with the `forget` output of load_tofu
+            retain_split = "retain" + str(100 - int(split.replace("forget", ""))).zfill(2)
+            self.retain_data =datasets.load_dataset(data_path, retain_split)["train"] # <<<< Replace with the `retain` output of load_tofu
+            self.model_configs = get_model_identifiers_from_yaml(model_family)
+            self.loss_type = loss_type
+
+            ...
+
```
philswatton: When putting code/outputs in markdown, make sure to mark it as code using backticks \` to enable syntax highlighting on GitHub

````suggestion
```
    Dataset({
        features: ['question', 'answer'],
        num_rows: N # N depends on the chosen forget fraction
    })
```

The first output is the forget set, and the second is the retain set.

It can (hopefully) be integrated into the source code in the following way, using the Text:

```python
    class TextForgetDatasetQA(Dataset):
        def __init__(self, data_path, tokenizer, model_family,  max_length=512, split = "forget10", loss_type="idk"):
            super(TextForgetDatasetQA, self).__init__()
            self.tokenizer = tokenizer
            self.max_length = max_length
            self.forget_data = datasets.load_dataset(data_path, split)["train"] # <<<< Replace with the `forget` output of load_tofu
            retain_split = "retain" + str(100 - int(split.replace("forget", ""))).zfill(2)
            self.retain_data =datasets.load_dataset(data_path, retain_split)["train"] # <<<< Replace with the `retain` output of load_tofu
            self.model_configs = get_model_identifiers_from_yaml(model_family)
            self.loss_type = loss_type

            ...
```
````

File: src/arcsf/data/data_utils.py [hunk 5]
```diff
@@ -0,0 +1,244 @@
...
+        rng.choice(
+            author_options, int(author_forget_fraction * n_authors), replace=False
+        )
+    )
+
+    return forget_authors
+
+
+def load_tofu(
+    granularity: str = "author_level",
```
philswatton: Suggesting splitting `granularity` into three arguments - `granularity`, `stratified`, and `forget_random` (with the last two taking `bool` inputs), one to handle granularity (so author or question) another flag for random vs deterministic forget set creation (implicitly, this will only affect question forgetting), and a third flag for stratified forgetting (again, implicitly will only affect question forgetting, and will interact with random).

In general, best to split up different conditions - if/when we add more this will help keep everything under control

src/arcsf/data/data_utils.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def get_indices_random(
+    within_author: bool = True,
+    forgotten_author_numbers: list = None,
+    q_remove_fraction: int = 0.1,
+    n_authors: int = 200,
+    q_per_author: int = 20,
+    random_seed: int = 42,
+):
```
philswatton: There's some of duplication between this and `get_indices_random`. Suggest extracting both to a single (set of) function(s) that remains agnostic to whether indices are selected randomly or not and handle that part separately in `load_tofu` (or in another fn), e.g.

EDIT - see below for recommendation on this

src/arcsf/data/data_utils.py
```diff
@@ -0,0 +1,244 @@
...
+
+    elif granularity == "structured_within_authors":
+        # biographical information generally contained in first 4 questions
+        forget_indices, retain_indices = get_indices_structured(
+            forget_author_numbers,
+            q_remove=qs_to_remove,
+            n_authors=author_count,
+            q_per_author=author_q_count,
+        )
+        debug_dict["qs_to_remove"] = qs_to_remove
```
philswatton: EDIT - see comment below instead of this one

philswatton (changes requested):
This is good work - much better than my first commit when I started here!

Some comments contained below/. The big ones are around refactoring data_utils.py - some of this code has some duplication and in general data_utils.py feels like it can be a little shorter. I've made some initial suggestions in this direction, happy to discuss during out meeting tomorrow.

Finally: I'd suggest renaming data_utils.py to load_data.py or even just tofu.py

src/arcsf/data/README.md
philswatton: Additionally: I'm not sure this README is required. For a single fn, the docstring should be enough. Certainly anything on the future direction of the project should be removed (e.g. hopefully fitting the tofu classes - this should stay in PRs/slack/discussion on the repo). README files in general should only be with a future user in mind

src/arcsf/data/data_utils.py
philswatton: Okay, ignore the above - I shouldn't write code recommendations near to 5pm. I'd suggest the following, using train_test_split from scikit-learn:

For getting forget numbers from question/author number:

```python
def flatten_list_of_lists(
    list: list[list]
) -> list:
    return sum(list, [])

def _get_forget_index(
    author: int,
    question: int,
    q_per_author: int,
) -> int:
    return question + q_per_author * author

def _get_forget_indices(
    authors: int,
    questions: int | list[int],
    q_per_author: int,
) -> int | list[int]:
    if isinstance(questions, int):
        return _get_forget_index(authors, questions, q_per_author)
    if isinstance(questions, list):
        return [_get_forget_index(authors, question, q_per_author) for question in questions]

def get_forget_indices(
    authors: int | list[int],
    questions: int | list[int],
    q_per_author: int,
) -> list[int]:
    if isinstance(authors, int):
        return _get_forget_indices(authors, questions, q_per_author)
    if isinstance(authors, list):
        out = [_get_forget_indices(author, questions, q_per_author) for author in authors]
        if isinstance(out[0], list):
            return flatten_list_of_lists(out)
        return out
```

For handling sampling:
```python
# Get full list of indices
all_indices = list(range(all_data.num_rows))

# Get full list of authors
authors = list(range(num_authors))

# Conditionally create full list of questions
if granularity == "author" or not forget_random:
    questions = list(range(q_per_author))

# Get indices
if granularity == "question":
    if forget_random:
        retain_indices, forget_indices = train_test_split(
            all_indices,
            test_size=num_forget,
            stratify=sum([[a]*q_per_author for a in authors], []) if stratify
            else None,
            random_state=random_seed,
        )
    if not forget_random:
        forget_indices = get_forget_indices(authors, questions, q_per_author)
        retain_indices = list(set(all_indices).difference(forget_indices))
if granularity == "author":
    retain_indices, forget_indices = train_test_split(
        authors,
        test_size=num_forget,
        random_state=random_seed,
    )
    retain_indices = get_forget_indices(retain_indices, questions, q_per_author)
    forget_indices = get_forget_indices(forget_indices, questions, q_per_author)

```

In general - if an existing function already does what you need, it's best to use this rather than write your own code (I should've started here yesterday!). The existing implementation will both be better than your own (as it'll have gone through far more rounds of development/more people) and makes life easier.

Edited to remove a couple of mistakes!

src/arcsf/data/data_utils.py
jack89roberts: The LoCoMoSeT docstring style is from the Google Python styleguide fyi: https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings (there are two main prevailing styles - the Google one and numpydoc, which is a bit more verbose). For reference this is how this docstring might look:

```suggestion
    """Returns Question--Answer indices given authors numbers to remove/retain, in the
    case where the forget target can be easily structured. Assumes equal questions per
    author, and sorted authors.

    Args:
        forgotten_author_numbers: Author indexes for which facts should be removed.
        n_authors: Total number of authors.
        q_removal: Number of questions/facts that should be removed.
        q_per_author: The number of questions to remove per author.

    Returns:
        A tuple (forget_indices, retain_indices) containing the indices of questions
        pertaining the authors in forgotten_author_numbers and the remaining question
        indices, respectively.
    """
```

Roughly changes are:

- Reducing indentation
- Reducing duplicated information about variable types (both after the variable name but also in it's description, e.g. we already know `forgotten_author_number` is a `list[int]` so its description doesn't need to start with that)
- Style is to treat the returns block as a single entity rather than breaking it into multiple vars.

jack89roberts: (and Copilot is pretty good at reducing the time taken to write docstrings btw)

src/arcsf/data/README.md
jack89roberts: Tend to agree this could be split into the `load_tofu` docstring + comments in issues (or new issues) about setting up data/model pipelines etc.

jack89roberts:
> Finally: I'd suggest renaming data_utils.py to load_data.py or even just tofu.py

I'd go `tofu.py`

src/arcsf/data/data_utils.py
jack89roberts: Still thinking about this but at least some of these values are TOFU constants and should be defined somewhere, e.g. `n_authors` and `q_per_author` I think will not change whenever using this with the original TOFU dataset so could be defined as `TOFU_N_AUTHORS`  and `TOFU_Q_PER_AUTHOR` and used throughout.

src/arcsf/data/data_utils.py
```diff
@@ -0,0 +1,244 @@
...
+
+        Parameters:
+                forgotten_author_numbers (list):    List of integers containing author
+                                                    indexes about which facts should
+                                                    be removed
+                n_authors (int):                    Total number of authors
+                q_removal (int):                    Number of questions/facts that
+                                                    should be removed
+                q_per_author (int):                 Integer defining the number of
+                                                    questions to remove per author
```
jack89roberts: Looking throughout I think this arg is always relating to the total number of questions per author in the TOFU dataset, not the number of questions to remove per author (which is `q_removal`)?

src/arcsf/data/data_utils.py
jack89roberts: Still thinking about this but at least some of these values are TOFU constants and should be defined somewhere, e.g. `n_authors` and `q_per_author` I think will not change whenever using this with the original TOFU dataset so could be defined as `TOFU_N_AUTHORS`  and `TOFU_Q_PER_AUTHOR` and used throughout.

src/arcsf/data/data_utils.py
```diff
@@ -0,0 +1,244 @@
...
+        Returns:
+                forget_indices (list):              List of integers corresponding to
+                                                    the question indices pertaining the
+                                                    authors in author_numbers
+                retain_indices (list):              List of integers corresponding to
+                                                    the remaining question indices
+    """
+    assert (
+        q_remove <= q_per_author
+    ), "Cannot remove more questions for an author than there are questions per author."
```
jack89roberts: Checks like this generally should be replaced with raising exceptions rather than using `assert` (as `assert` can be ignored at runtime)

src/arcsf/data/data_utils.py
```diff
@@ -0,0 +1,244 @@
...
+    indices_list = list()
```
jack89roberts: minor but generally `[]` is preferred

jack89roberts:
On all the comments re indexing etc. I don't think we need to reinvent the wheel if something already works. But my main suggestion is to consider whether it would be helpful to create `author_id` and `question_id` columns in the dataset at the beginning, which could also be used in the various sampling/filtering functions. One pro of this is it might make some analysis easier later - e.g. we wouldn't need to keep propagating information about which questions/authors have been removed to figure out which rows in the retain/forget sets correspond to which authors (as we could just refer to the `author_id`).

HuggingFace `Datasets` also have some in-built functions that could help: https://huggingface.co/docs/datasets/en/process . E.g. with an `author_id` column the `author_level` split could just be something like (probably not the most efficient way):

```python
forget_set = all_data.filter(lambda row: row["author_id"] in forgotten_author_numbers)
retain_set = all_data.filter(lambda row: row["author_id"] not in forgotten_author_numbers)
```

philswatton:
> On all the comments re indexing etc. I don't think we need to reinvent the wheel if something already works. But my main suggestion is to consider whether it would be helpful to create `author_id` and `question_id` columns in the dataset at the beginning, which could also be used in the various sampling/filtering functions. One pro of this is it might make some analysis easier later - e.g. we wouldn't need to keep propagating information about which questions/authors have been removed to figure out which rows in the retain/forget sets correspond to which authors (as we could just refer to the `author_id`).
> 
> HuggingFace `Datasets` also have some in-built functions that could help: https://huggingface.co/docs/datasets/en/process . E.g. with an `author_id` column the `author_level` split could just be something like (probably not the most efficient way):
> 
> ```python
> forget_set = all_data.filter(lambda row: row["author_id"] in forgotten_author_numbers)
> retain_set = all_data.filter(lambda row: row["author_id"] not in forgotten_author_numbers)
> ```

I was thinking about this too - am open to arguments either way for now, but we'll almost certainly want to have our own indexing for each level of hierarchy when we start generating our own data in #2

src/arcsf/data/data_utils.py
philswatton: Agreed re extraction to constants - happy for these to be used in main `load_tofu` fn but suggest leaving rest w/out defaults in case we use them when extending the data in #2

src/arcsf/data/tofu.py
```diff
@@ -52,20 +59,41 @@ def load_tofu(
     forgotten_fact_fraction: float,
     random_seed: int,
     debug=False,
-) -> tuple[Dataset, Dataset, dict]:
+) -> tuple[Dataset, Dataset, dict] | tuple[Dataset, Dataset]:
+    """
+    Loads TOFU dataset given different flags for retain--forget split.
+    Params:
```
jack89roberts, line 67: `Args:` rather than `Params:`

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,194 @@
...
+    stratified: bool,
+    forget_random: bool,
+    forgotten_author_fraction: float,
+    forgotten_fact_fraction: float,
+    random_seed: int,
+    debug=False,
+) -> tuple[Dataset, Dataset, dict] | tuple[Dataset, Dataset]:
+    """
+    Loads TOFU dataset given different flags for retain--forget split.
+    Params:
```
jack89roberts:
```suggestion
    Args:
```

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,194 @@
+import math
+
+import numpy as np
+from datasets import Dataset, load_dataset
+from sklearn.model_selection import train_test_split
+
+
+# self explanatory
```
jack89roberts: Delete

```suggestion
```

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,194 @@
...
+def _flatten_list_of_lists(list: list[list]) -> list:
+    return sum(list, [])
+
+
+# returns the question index in the dataset, given an author number and question
+# number (within the author) (inner function)
```
jack89roberts: move comments like these to a docstring inside the function

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+
+    debug_dict["forget_author_numbers"] = forget_authors
+
+    if granularity == "question":
+        if forget_random:
+            if stratified:
+                # if stratified and random: select train-test split within each
+                # set of author questions (currently selects same indices
+                # TODO: Is this the behaviour we want?
```
jack89roberts: Rather than TODO make an issue, or link out to an issue here

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,193 @@
+import math
+
+import numpy as np
+from datasets import Dataset, load_dataset
+from sklearn.model_selection import train_test_split
```
jack89roberts, line 6: scikit-learn should be in pyproject.toml if we're using it (we get it installed accidentally as a 3rd party dependency of something else at the moment)

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,193 @@
...
+    return sum(list, [])
+
+
+# returns the question index in the dataset, given an author number and question
+# number (within the author) (inner function)
+def _get_forget_index(
+    author: int,
+    question: int,
+    q_per_author: int,
+) -> int:
```
jack89roberts: Short enough to fit on one line:

```suggestion
def _get_forget_index(author: int, question: int, q_per_author: int) -> int:
```

also some other functions in this file (black will refactor them if you remove the comma after the last arg)

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,193 @@
...
+    authors: int,
+    questions: int | list[int],
+    q_per_author: int,
+) -> int | list[int]:
+    if isinstance(questions, int):
+        return _get_forget_index(authors, questions, q_per_author)
+    if isinstance(questions, list):
+        return [
+            _get_forget_index(authors, question, q_per_author) for question in questions
+        ]
```
jack89roberts: The 2nd `if` isn't really needed (if it's not an int it's required to be a list per the docstring):

```suggestion
    if isinstance(questions, int):
        return _get_forget_index(authors, questions, q_per_author)

    return [
        _get_forget_index(authors, question, q_per_author) for question in questions
    ]
```

Alternatively, if you do want to check the type it should raise an error if it's neither of those types (at the moment it would just get to the end of the function and return None)

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,193 @@
...
+) -> list[int]:
+    if isinstance(authors, int):
+        return _get_forget_indices(authors, questions, q_per_author)
+    if isinstance(authors, list):
+        out = [
+            _get_forget_indices(author, questions, q_per_author) for author in authors
+        ]
+        if isinstance(out[0], list):
+            return _flatten_list_of_lists(out)
+        return out
```
jack89roberts: as per `_get_forget_indices`:

```suggestion
    if isinstance(authors, int):
        return _get_forget_indices(authors, questions, q_per_author)

    out = [
        _get_forget_indices(author, questions, q_per_author) for author in authors
    ]
    if isinstance(out[0], list):
        return _flatten_list_of_lists(out)
    return out
```

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,193 @@
...
+
```
jack89roberts: Do both these functions boil down to:

```
if isinstance(authors, int):
    authors = [authors]
if isinstance(questions, int):
    questions = [questions]

return [_get_forget_index(author, question, q_per_author) for author in authors for question in questions]
```

?

My original comment was going to be that maybe `get_forget_indices` and `_get_forget_indices` should be renamed to reflect what they're looping over, e.g. perhaps `_get_forget_indices(authors, questions, q_per_author)` should be `_get_author_question_indices(author, questions, q_per_author)` (also note `author` instead of `authors` since it's expected to get a single author id as input). But based on the above I'm not sure they need to be separated into two functions at all.

There's also the option of filtering based on the author/question id columns you've created.

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,193 @@
...
+    # Get full list of indices
+    all_indices = list(range(all_data.num_rows))
+
+    # Get full list of authors
+    authors = list(range(num_authors))
+
+    # Conditionally create full list of questions
+    if granularity == "author" or not forget_random:
+        forget_questions = list(range(num_forget))
+
```
jack89roberts, line 95: this doesn't seem to be used again until line 172, can it be moved closer to there to help understand what it's for?

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,193 @@
...
+        if forget_random:
+            if stratified:
+                # if stratified and random: select train-test split within each
+                # set of author questions (currently selects same indices
+                # TODO: Is this the behaviour we want?
+                forget_indices = []
+                for author in forget_authors:
+                    all_author_indices = np.arange(
+                        get_forget_indices(author, 0, q_per_author),
+                        get_forget_indices(author, 20, q_per_author),
```
jack89roberts: Remove hardcoded 20?

```suggestion
                        get_forget_indices(author, q_per_author, q_per_author),
```

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,193 @@
...
+                    test_size=num_forget,
+                    random_state=random_seed,
+                )
+                # need to redefine the forgotten authors now - can use np.floor
+                debug_dict["forget_author_numbers"] = np.floor(
+                    np.array(forget_indices) / 20
+                ).tolist()
+        # if not random then take first num_forget questions
+        # forget_questions contains the relevant indices for this
+        if not forget_random:
```
jack89roberts:
```suggestion
        else:
```

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,193 @@
...
+            forget_indices = get_forget_indices(
+                forget_authors, forget_questions, q_per_author
+            )
+            retain_indices = list(set(all_indices).difference(forget_indices))
+
+    # if granularity == author we just use retain/forget authors to define our splits
+    if granularity == "author":
```
jack89roberts:
```suggestion
    else:
```

or `elif granularity == "author"` (but question and author are the only options)

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,193 @@
...
+                num_forget = int(q_per_author * forgotten_fact_fraction)
+        # non stratified random: forgotten_fact_fraction applied across entire dataset
+        elif not stratified:
+            num_forget = int(num_authors * q_per_author * forgotten_fact_fraction)
+
+    debug_dict = {
+        "author_count": num_authors,
+        "q_per_author": q_per_author,
+        "num_forget": num_forget,
+    }
```
jack89roberts: would be nicer if the tests can be written to not require this

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,193 @@
...
+        - Two datasets with forget and retain sets
+            as well as a debugging dictionary (optional)
+    """
+
+    all_data = load_dataset("locuslab/TOFU", "full")["train"]
+
+    num_authors = 200  # hard coding author count for now
+    q_per_author = 20  # hard coding author question count for now
+
+    q_indices = [q for q in range(num_authors * q_per_author)]
```
jack89roberts, line 64: not sure but maybe should be a per-author index, e.g. author 0 question 0, ... author 0 question 19, author 1 question 0, author 19 question 19, ... etc.? Or have both possibly (an overall row index and a question index for which number question it is for that author)

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,193 @@
...
+            else:
+                retain_indices, forget_indices = train_test_split(
+                    all_indices,
+                    test_size=num_forget,
+                    random_state=random_seed,
+                )
+                # need to redefine the forgotten authors now - can use np.floor
+                debug_dict["forget_author_numbers"] = np.floor(
+                    np.array(forget_indices) / 20
+                ).tolist()
```
jack89roberts: is it right that this is nested in the else? (is it not needed in other cases?)

src/arcsf/data/tofu.py
```diff
@@ -0,0 +1,193 @@
...
+                for author in forget_authors:
+                    all_author_indices = np.arange(
+                        get_forget_indices(author, 0, q_per_author),
+                        get_forget_indices(author, 20, q_per_author),
+                    )
+                    _, author_forget_indices = train_test_split(
+                        all_author_indices,
+                        test_size=num_forget,
+                        random_state=random_seed,
+                    )
```
jack89roberts, line 118: `train_test_split` has a `stratify` arg that could simply this/maybe allow the if/else stratified blocks to be merged into one. E.g. something like `train_test_split(all_indices, author_indices, test_size=num_forget, random_state=random_seed, stratify=True)`

src/arcsf/data/tofu.py
J-Dymond: Fixed this by adding through poetry

src/arcsf/data/tofu.py
jack89roberts: ignore it is used earlier

src/arcsf/data/tofu.py
jack89roberts: didn't work for other reasons

src/arcsf/data/tofu.py
jack89roberts: ok

J-Dymond: just because it gets updated

src/arcsf/data/tofu.py
J-Dymond: changed

jack89roberts (approved):
Good to merge pending the last few changes we discussed.

src/arcsf/data/tofu.py
J-Dymond: I think I will keep it like this for now

src/arcsf/data/tofu.py
J-Dymond: Done

src/arcsf/data/tofu.py
J-Dymond: Fixed

src/arcsf/data/tofu.py
J-Dymond: Changed this behaviour

src/arcsf/data/tofu.py
J-Dymond: They might do, will investigate this today

src/arcsf/data/tofu.py
J-Dymond: Debug dict removed, defined in tests. Had to remove a test that checked the dict against output, but adding Q+A indices makes this irrelevant.

src/arcsf/data/tofu.py
J-Dymond: This didn't work, so I renamed functions to make them clearer

PR #13, opened 2024-04-17
7 implement data preprocessing

### LINKED ISSUE #7: Implement Data Preprocessing

J-Dymond:
Initial implementation of the data_modules code from the tofu repo. Currently I test them using an identity function for the tokenizer and in some cases the formatter. There are three, one for returning QA pairs, one for finetuning which returns a formatted Question--Answer pair for autoregression, and one for a forgetting script which returns a formatted retain example and a formatted forget sample. In all cases the forget answer can be replaced with an "idk" type phrase, which is randomly sampled from a .jsonl file.

### DIFF
--- a/src/arcsf/data/data_module.py
+++ b/src/arcsf/data/data_module.py
@@ -4,14 +4,31 @@ from torch.utils.data import ConcatDataset, Dataset
 from arcsf.data.data_utils import load_tofu
 
 
-def QAformatter_basic(QA):
+def QAformatter_basic(QA: tuple[str]) -> str:
+    """
+    Basic QA formatter which accepts a tuple outputs:
+
+    "Question: [input question]\\nAnswer: [input answer]"
+
+    args:
+        - QA: Tuple of question answer pair
+    returns:
+        - full_text: formatted question--answer pair
+    """
     Question, Answer = QA
     question_formatting = "Question: " + Question
     answer_formatting = "\nAnswer: " + Answer
-    return question_formatting + answer_formatting
+    full_text = question_formatting + answer_formatting
+    return full_text
 
 
 class QADataSet(Dataset):
+    """
+    Question answer format dataset, __getitem__ returns a question--answer pair as
+    a tuple. There is an option to output the answers using "I don't know" synonyms
+    by specifying loss_type as "idk".
+    """
+
     def __init__(
         self,
         tokenizer,
@@ -58,6 +75,13 @@ class QADataSet(Dataset):
 
 
 class FinetuneDataset(Dataset):
+    """
+    Finetune version of the dataset, __getitem__ returns a sample taken either from
+    retain, forget subsets, or a combination of both. Samples are formatted using a
+    question formatter allowing for autoregression. There is an option to output
+    samples using "I don't know" synonyms by specifying loss_type as "idk".
+    """
+
     def __init__(
         self,
         tokenizer,
@@ -110,6 +134,12 @@ class FinetuneDataset(Dataset):
 
 
 class QAForgetDataSet(Dataset):
+    """
+    Q+A Forget version of the dataset, __getitem__ returns a retain and forget sample.
+    Both are formatted using a question formatter. There is an option to output samples
+    using "I don't know" synonyms by specifying loss_type as "idk".
+    """
+
     def __init__(
         self,
         tokenizer,
@@ -119,18 +149,20 @@ class QAForgetDataSet(Dataset):
         q_to_drop=0.1,
         loss_type="standard",
         random_seed=42,
+        debug=False,
     ):
         super(QAForgetDataSet, self).__init__()
         self.tokenizer = tokenizer
         self.qa_formatter = qa_formatter
         self.loss_type = loss_type
+        self.debug = debug
 
         self.forget_data, self.retain_data, self.debug_dict = load_tofu(
             granularity,
             forgotten_author_fraction=a_to_drop,
             forgotten_fact_fraction=q_to_drop,
             random_seed=random_seed,
-            debug=True,
+            debug=debug,
         )
         # shuffle the retain data and get the question indices for debugging
         self.retain_data = self.retain_data.shuffle(seed=random_seed)
@@ -146,8 +178,6 @@ class QAForgetDataSet(Dataset):
         return len(self.data)
 
     def __getitem__(self, idx):
-        # TODO: check team are happy with this over random sampling
-        # TODO: investigate random reshuffle when finished
         # this takes the first item in our retain data permutation using item_index
         retain_row = self.retain_data[self.item_index[0]]
         # then rolls the permutation vector using the item_index to ensure
@@ -170,10 +200,13 @@ class QAForgetDataSet(Dataset):
         retain = self.qa_formatter((retain_question, retain_answer))
         forget = self.qa_formatter((forget_question, forget_answer))
 
-        return (
-            self.tokenizer(retain),
-            retain_row["question_index"],
-        ), (
-            self.tokenizer(forget),
-            forget_row["question_index"],
-        )
+        if self.debug:
+            return (
+                self.tokenizer(retain),
+                retain_row["question_index"],
+            ), (
+                self.tokenizer(forget),
+                forget_row["question_index"],
+            )
+        else:
+            return self.tokenizer(retain), self.tokenizer(forget)
--- a/src/arcsf/data/data_module.py
+++ b/src/arcsf/data/data_module.py
@@ -6,6 +6,28 @@ from arcsf.data.tofu import load_tofu
 _dataset_dict = {"tofu": load_tofu}
 
 
+def get_data(
+    dataset_name,
+    granularity,
+    stratified,
+    forget_random,
+    forgotten_author_fraction,
+    forgotten_fact_fraction,
+    random_seed,
+):
+    load_func = _dataset_dict[dataset_name]
+    data = load_func(
+        granularity,
+        stratified,
+        forget_random,
+        forgotten_author_fraction,
+        forgotten_fact_fraction,
+        random_seed=random_seed,
+    )
+
+    return data
+
+
 def QAformatter_basic(QA: tuple[str]) -> str:
     """
     Basic QA formatter which accepts a tuple outputs:
@@ -33,34 +55,16 @@ class QADataSet(Dataset):
 
     def __init__(
         self,
-        dataset_name,
+        data,
         tokenizer,
         qa_formatter,
-        granularity,
-        stratified,
-        forget_random,
         split,
-        a_to_drop,
-        q_to_drop,
         loss_type,
-        random_seed=42,
     ):
         super(QADataSet, self).__init__()
         self.tokenizer = tokenizer
         self.qa_formatter = qa_formatter
         self.loss_type = loss_type
-
-        get_data = _dataset_dict[dataset_name]
-
-        data = get_data(
-            granularity,
-            stratified,
-            forget_random,
-            forgotten_author_fraction=a_to_drop,
-            forgotten_fact_fraction=q_to_drop,
-            random_seed=random_seed,
-        )
-
         forget_data, retain_data = data
 
         split_dict = {"retain": retain_data, "forget": forget_data}
@@ -69,6 +73,16 @@ class QADataSet(Dataset):
         if loss_type == "idk":
             with open("src/arcsf/data/idk.jsonl") as idk_file:
                 self.idk = idk_file.read().splitlines()
+            self.answer_sampler = self.idk_sampler
+        else:
+            self.answer_sampler = self.forget_sampler
+
+    def idk_sampler(self, _):
+        rand_pos = torch.randint(0, len(self.idk), (1,)).item()
+        return self.idk[rand_pos]
+
+    def forget_sampler(self, forget_row):
+        return forget_row["answer"]
 
     def __len__(self):
         return len(self.data)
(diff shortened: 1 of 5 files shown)


jack89roberts (commented):
~~Left a few suggestions to talk through but looking good~~ oops wrong PR

src/arcsf/data/data_module.py
```diff
@@ -0,0 +1,207 @@
...
+def QAformatter_basic(QA: tuple[str]) -> str:
+    """
+    Basic QA formatter which accepts a tuple outputs:
+
+    "Question: [input question]\nAnswer: [input answer]"
+
+    args:
+        - QA: Tuple of question answer pair
+    returns:
+        - full_text: formatted question--answer pair
```
jack89roberts: Just for convention/consistency, if writing full docstrings the Googledoc style is:

```suggestion
    Basic QA formatter which accepts a tuple and outputs:
    "Question: [input question]\nAnswer: [input answer]"

    Args:
        qa: Tuple of question answer pair
    
    Returns:
        Formatted question--answer pair
```

(Capitalise `Args` and `Returns`, no `-` before args/returns, whitespace between Args and Returns blocks, returns doesn't need a variable name).

src/arcsf/data/data_module.py
```diff
@@ -0,0 +1,207 @@
...
+    synonyms by specifying loss_type as "idk".
+    """
+
+    def __init__(
+        self,
+        data,
+        tokenizer,
+        qa_formatter,
+        split,
+        loss_type,
```
jack89roberts: types and docstring to explain what these are expected to be would be helpful, e.g. the possible values for `loss_type` and what that changes. Same for the `__init__` of `FinetuneDataset` and `QAForgetDataset`

src/arcsf/data/data_module.py
```diff
@@ -0,0 +1,207 @@
...
+
+        question = self.data[idx]["question"]
+        answer = self.qa_formatter(self.data[idx]["answer"])
+
+        inp = self.qa_formatter((question, answer))
+
+        return self.tokenizer(inp)
+
+
+class QAForgetDataSet(Dataset):
```
jack89roberts: DataSet -> Dataset for consistency (likely needs updating elsewhere too)

```suggestion
class QAForgetDataset(Dataset):
```

src/arcsf/data/data_module.py
```diff
@@ -0,0 +1,207 @@
...
+        self.qa_formatter = qa_formatter
+        self.loss_type = loss_type
+        forget_data, retain_data = data
+
+        split_dict = {"retain": retain_data, "forget": forget_data}
+        self.data = split_dict[split]
+
+        if loss_type == "idk":
+            with open("src/arcsf/data/idk.jsonl") as idk_file:
+                self.idk = idk_file.read().splitlines()
```
jack89roberts: This path depends on being in the parent `arc-selective-forgetting` directory (i.e. `src/arcsf/...` is a relative path). I usually use `__file__` to get around this (which gives the absolute path to the current file), which there's an example of for the definition of `TEST_DATA_DIR` in `conftest.py`. But the recommended way seems to be to use `importlib.resources` now, see https://stackoverflow.com/a/20885799

src/arcsf/data/data_module.py
```diff
@@ -0,0 +1,207 @@
...
+            self.answer_sampler = self.idk_sampler
+        else:
+            self.answer_sampler = self.forget_sampler
+
+    def answer_sampler():
+        return None
+
+    def idk_sampler(self, _):
+        rand_pos = torch.randint(0, len(self.idk), (1,)).item()
```
jack89roberts: To make this reproducible it should be possible to set the initial random seed for this (e.g. by making a random number generator when the class is created and using it here - not sure how that works in `torch` exactly but in `numpy` you can instantiate a rng).

src/arcsf/data/data_module.py
```diff
@@ -0,0 +1,207 @@
...
+
+        if loss_type == "idk":
+            with open("src/arcsf/data/idk.jsonl") as idk_file:
+                self.idk = idk_file.read().splitlines()
+            self.answer_sampler = self.idk_sampler
+        else:
+            self.answer_sampler = self.forget_sampler
+
+    def idk_sampler(self, _):
+        rand_pos = torch.randint(0, len(self.idk), (1,)).item()
```
jack89roberts: See earlier comment re random seeds

src/arcsf/data/data_module.py
```diff
@@ -0,0 +1,207 @@
...
+
+        if loss_type == "idk":
+            with open("src/arcsf/data/idk.jsonl") as idk_file:
+                self.idk = idk_file.read().splitlines()
+            self.answer_sampler = self.idk_sampler
+        else:
+            self.answer_sampler = self.forget_sampler
+
+    def answer_sampler():
+        return None
```
jack89roberts: this was us messing about earlier, if you want to keep it should raise a `NotImplementedError`

src/arcsf/data/data_module.py
```diff
@@ -0,0 +1,207 @@
...
+    Basic QA formatter which accepts a tuple outputs:
+
+    "Question: [input question]\nAnswer: [input answer]"
+
+    args:
+        - QA: Tuple of question answer pair
+    returns:
+        - full_text: formatted question--answer pair
+    """
+    Question, Answer = QA
```
jack89roberts: Functions and variables/args should be lowercase by convention, e.g.

- `QAformatter_basic` -> `qa_formatter_basic`
- `QA` -> `qa`
- `Question` -> `question`
- `Answer` -> `answer`

src/arcsf/data/data_module.py
```diff
@@ -0,0 +1,207 @@
...
+    question_formatting = "Question: " + Question
+    answer_formatting = "\nAnswer: " + Answer
+    full_text = question_formatting + answer_formatting
+    return full_text
```
jack89roberts: Could be a lot shorter using an f-string:

```suggestion
    return f"Question: {question}\nAnswer: {answer}"
```

src/arcsf/data/data_module.py
```diff
@@ -0,0 +1,207 @@
...
+
+        return self.tokenizer(input), self.tokenizer(target)
+
+
+class FinetuneDataset(Dataset):
+    """
+    Finetune version of the dataset, __getitem__ returns a sample taken either from
+    retain, forget subsets, or a combination of both. Samples are formatted using a
+    question formatter allowing for autoregression. There is an option to output
+    samples using "I don't know" synonyms by specifying loss_type as "idk".
```
jack89roberts: `idk` loss doesn't apply for fine-tuning

src/arcsf/data/data_module.py
```diff
@@ -0,0 +1,207 @@
...
+            self.answer_sampler = self.forget_sampler
+
+    def answer_sampler():
+        return None
+
+    def idk_sampler(self, _):
+        rand_pos = torch.randint(0, len(self.idk), (1,)).item()
+        return self.idk[rand_pos]
+
+    def forget_sampler(self, forget_row):
```
jack89roberts: Is `forget_sampler` the right name for this, will it always be a forget row (e.g. what is split is retrain)? `get_answer` maybe?

src/arcsf/data/data_module.py
```diff
@@ -0,0 +1,207 @@
...
+        self.qa_formatter = qa_formatter
+
+        forget_data, retain_data = data
+
+        split_dict = {
+            "retain": retain_data,
+            "forget": forget_data,
+            "all": ConcatDataset([retain_data, forget_data]),
+        }
+        self.data = split_dict[split]
```
jack89roberts: `FinetuneDataset` and `EvalQADataset` both only use a single data split (at least at the moment), would it be easier if they were created only with the data they need (so the data argument was a single dataset not a tuple of forget and retain)? That would then also avoid needing this dict/concat step here, which feels a bit odd (start with whole dataset, split it, combine it again).

src/arcsf/data/data_module.py
```diff
@@ -0,0 +1,207 @@
...
+        # forget data works as in above examples
+        forget_row = self.forget_data[idx]
+        forget_question = forget_row["question"]
+
+        forget_answer = self.answer_sampler(forget_row)
+
+        retain = self.qa_formatter((retain_question, retain_answer))
+        forget = self.qa_formatter((forget_question, forget_answer))
+
+        return self.tokenizer(retain), self.tokenizer(forget)
```
jack89roberts, line 255: There will likely need to be more logic than only calling the tokenizer here and in the other classes, e.g. see `convert_raw_data_to_model_format` in the TOFU code, but it may be easier to figure out when integrating this with the fine-tunung/forgetting code.

jack89roberts (changes requested):

src/arcsf/data/data_module.py
J-Dymond: Sorry, I will fix these!

src/arcsf/data/data_module.py
J-Dymond: Will make these changes on all of the classes. Wasn't sure on the convention for docstrings in classes.

src/arcsf/data/data_module.py
J-Dymond: Will make this for the three dataset lasses

src/arcsf/data/data_module.py
J-Dymond: Perfect will investigate this

src/arcsf/data/data_module.py
J-Dymond: done

src/arcsf/data/data_module.py
J-Dymond: Will remove this reference

src/arcsf/data/data_module.py
J-Dymond: Yeah I thought so, this was intended as more of a placeholder, until we know what we want it to do. I thought qa_formatter may be able to handle this logic? as in the TOFU code.

src/arcsf/data/data_module.py
J-Dymond: have done, created a generator for the class to use

src/arcsf/data/data_module.py
J-Dymond: done

src/arcsf/data/data_module.py
J-Dymond: Ok, will investigate this

J-Dymond: Addressed this now here and in tests, though `load_tofu` outputs a tuple, so its not necessarily obvious which split we are passing to the class using the index. Also these will now operate differently to the `ForgetDataset`. But I suppose this can be addressed down the line.

tests/test_data_module.py
```diff
@@ -0,0 +1,119 @@
...
+    """Check that when using an idk loss, that the targets are correct."""
+    data = get_data(
+        "tofu",
+        granularity="question",
+        stratified=False,
+        forget_random=True,
+        forgotten_author_fraction=1 / 3,
+        forgotten_fact_fraction=1 / 3,
+        random_seed=42,
+    )[0]
```
jack89roberts: Unless I missed something the `get_data` call here can be replaced with the `data` fixture created earlier:

```suggestion
def test_idk_targets(data):
    """Check that when using an idk loss, that the targets are correct."""
```

tests/test_data_module.py
```diff
@@ -0,0 +1,119 @@
...
+    idk_set = EvalQADataset(
+        data,
+        _identity,
+        _identity,
+        loss_type="idk",
+    )
+    with open("src/arcsf/data/idk.jsonl") as idk_file:
```
jack89roberts: path issue in previous comment also applies here, maybe worth having a `get_idk_responses` function in `arcsf/data` somewhere

tests/test_data_module.py
J-Dymond: Yeah, I think you're right. Made this change

src/arcsf/data/data_module.py
J-Dymond: Thank you for investigating this...

tests/test_data_module.py
J-Dymond: You looked at this for me, thank you

src/arcsf/data/data_module.py
J-Dymond: Have removed the `get_answer` placeholder now, so the code is cleaner

src/arcsf/data/data_module.py
```diff
@@ -62,12 +81,25 @@ class EvalQADataset(Dataset):
 
     def __init__(
         self,
-        data,
-        tokenizer,
-        qa_formatter,
-        loss_type,
+        data: Dataset,
+        tokenizer: AutoTokenizer,
+        qa_formatter: callable,
```
jack89roberts: Instead of `callable` (which is a built-in python function) the typehint should use `Callable` from `collections.abc`: https://docs.python.org/3/library/typing.html#annotating-callable-objects

(i.e. do `from collections.abc import Callable` and then replace all `callable` with `Callable`. You can also type hint the args and return type of the function if you want).

jack89roberts (approved):

src/arcsf/data/data_module.py
J-Dymond: Thank you, I wasn't sure how to typehint a function, I've amended this now.

PR #14, opened 2024-04-18
8 fine tuning

### LINKED ISSUE #8: Add fine-tuning to pipeline
- Should be a fairly straightforward implementation of fine-tuning via HF transformers library
- Leave open option for PEFT via LoRA (+ quantisation?)

philswatton:
UPDATED: A PR to resolve #8, now open for review.

Big picture:

- Implements model fine-tuning
- Implements a partial config structure for the project, with a view towards minimal changes being required to cover the full project pipeline
- Implements Baskerville array jobs\
- Implements wandb usage

**The code here:**

- Adds functions for loading model, tokenizer, and trainer
- Adds a base config class, along with experiment, data, and model config classes
- The idea is to use a modular setup for configs, with one config per experiment component (data, model, forgetting techniques) to allow us to easily change different components of the code without needing to change all components
  - I anticipate being able to add forget-specific fine-tuning details to model hyperparam configs
- A function that given a fairly simple top level experiment config, spawns several single experiment configs (and, optionally, an associated baskerville array script)
- Example experiment (both normal and top-level), data, model, and hyperparam configs
- scripts/README containing instructions for fine-tuning
- Some additional dependencies
- Unit tests


src/arcsf/config/config_class.py
```diff
@@ -0,0 +1,49 @@
...
+
+class Config(ABC):
+
+    def __init__(
+        self,
+    ) -> None:
+        super().__init__()
+
+    @abstractclassmethod
+    def from_dict(cls, dict) -> "Config":
```
jack89roberts:
```suggestion
from abc import ABC, abstractmethod

import yaml


class Config(ABC):

    def __init__(
        self,
    ) -> None:
        super().__init__()

    @abstractmethod
    @classmethod
    def from_dict(cls, dict) -> "Config":
```

src/arcsf/config/config_class.py
jack89roberts, line 1: Minor but I prefer trying to keep names short for imports e.g. just `experiment.py` rather than `experiment_config.py`, and also `config_class.py` in `models/config_class.py` (-> `models/config.py`). For this one `config/config.py` would be a bit weird but maybe stick with that to be consistent.

src/arcsf/models/config_class.py
```diff
@@ -0,0 +1,62 @@
...
+        # TODO: work out how to manage this in line w/ varying dataset size
+        self.logging_strategy = trainer_kwargs["logging_strategy"]
+        self.logging_dir = f'{trainer_kwargs["output_dir"]}/logs'
+        self.logging_steps = trainer_kwargs["logging_steps"]
+
+        # Early stopping - TODO make optional
+        self.save_strategy = trainer_kwargs["save_strategy"]
+        self.save_steps = trainer_kwargs["logging_steps"]
+        self.load_best_model_at_end = trainer_kwargs["load_best_model_at_end"]
+        self.metric_for_best_model = trainer_kwargs["metric_for_best_model"]
```
jack89roberts: For me it's probably overkill to explicitly define args for things we'll be passing straight to `TrainingArguments` when there's likely to be quite a few things we'll want to adjust and `TrainingArguments` has a lot of other defaults/args that we're not controlling for here anyway. I would lean towards just keeping the dict of trainer kwargs and making sure we have all the settings in the eventual `TrainingArguments` instance logged (which I think `wandb` does anyway).

Some of them, like `output_dir`, it may make sense to have defined separately in the class for other purposes too though.

src/arcsf/config/experiment_config.py
```diff
@@ -0,0 +1,103 @@
...
+            environment = Environment(loader=FileSystemLoader("src/arcsf/config"))
+            template = environment.get_template("jobscript_template.sh")
+
+            script_content = template.render(
+                job_name=f"submit_{n}",
+                walltime=top_config["bask"]["walltime"],
+                node_number=top_config["bask"]["node_number"],
+                gpu_number=top_config["bask"]["gpu_number"],
+                cpu_per_gpu=top_config["bask"]["cpu_per_gpu"],
+                # array_number=, #TODO
```
jack89roberts: Just `n` I think? Ah but would need the relevant bit in the jobscript too.

src/arcsf/config/experiment_config.py
```diff
@@ -0,0 +1,103 @@
...
+                script_name="scripts/train.py",
+                experiment_file=file_name,
+            )
+
+            with open(f"{baskdir}/submit_{n}.sh", "w") as f:
+                f.write(script_content)
+
+
+class ExperimentConfig(Config):
```
jack89roberts: Consider we'll probably have fine-tuning, forgetting, and evaluating jobs (so maybe bits of this are a base class and others are split into a fine-tuning class).

Also in general not sure whether following the Experiment/top level config structure in LoCoMoSeT is the best way to structure everything tbh. But I'd focus on getting the single config stuff working first anyway, before worrying too much about being able to scan over parameters etc.

src/arcsf/models/model.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def load_model_and_tokenizer(
+    model_id: str,
+    peft_kwargs: dict | None = None,
+    **model_kwargs,
+) -> tuple[PreTrainedModel | PeftModel, PreTrainedTokenizer]:
+
+    # Initialise
+    add_token_to_model = True
```
jack89roberts: Safer for this to start False?

```suggestion
    add_token_to_model = False
```

src/arcsf/models/model.py
```diff
@@ -0,0 +1,46 @@
...
+
+    # Load tokenizer
+    tokenizer = AutoTokenizer.from_pretrained(model_id)
+
+    # Optionally add padding token
+    if tokenizer.pad_token is None:
+        tokenizer.add_special_tokens({"pad_token": "<|padding|>"})
+        add_token_to_model = True
```
jack89roberts: This maybe needs to be model config dependent, may not always be the right thing to do.

src/arcsf/models/trainer.py
```diff
@@ -0,0 +1,65 @@
...
+        logging_strategy=config.logging_strategy,
+        logging_dir=config.logging_dir,
+        logging_steps=config.logging_steps,
+        # Early stopping - TODO make optional
+        save_strateg=config.save_strategy,
+        save_steps=config.save_steps,
+        load_best_model_at_end=config.load_best_model_at_end,
+        metric_for_best_model=config.metric_for_best_model,
+        # TODO add seed, consider other args
+    )
```
jack89roberts: Would change if not explicitly defining everything in the config class.

src/arcsf/__init__.py
```diff
@@ -0,0 +1,6 @@
+from arcsf.config.experiment_config import (  # noqa F401
+    ExperimentConfig,
+    generate_experiment_configs,
+)
+from arcsf.models.model import load_model_and_tokenizer  # noqa F401
+from arcsf.models.trainer import load_trainer  # noqa F401
```
jack89roberts: It makes some imports shorter but is this worth it? Or if you want to keep it then you can use `__all__` rather than relying on Flake8 ignores:


```suggestion
from arcsf.config.experiment_config import ExperimentConfig, generate_experiment_configs
from arcsf.models.model import load_model_and_tokenizer
from arcsf.models.trainer import load_trainer

__all__ = [
    "ExperimentConfig",
    "generate_experiment_configs",
    "load_model_and_tokenizer",
    "load_trainer",
]
```

jack89roberts (commented):
Few initial thoughts/comments

src/arcsf/models/config_class.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class ModelConfig(Config):
+
+    def __init__(
+        self,
+        model_id: str,
+        model_kwargs: dict,
+        trainer_kwargs: dict,
+    ) -> None:
```
jack89roberts: Thinking about this in terms of making it flexible for unlearning as well, maybe this could be modified to take a `trainer_type`, I guess as a `str` which can be used to load the relevant trainer/forgetter class (the draft forgetting PR has a class for each forget method which are all variants of the HuggingFace trainer with a different loss function and maybe extra arguments).

That could be done in the forgetting PR, though also worth thinking about config structure as fine-tuning and unlearning will likely need different hyperparameters. E.g. in the TOFU repo they have separate [model](https://github.com/locuslab/tofu/blob/main/config/model_config.yaml), [fine-tuning](https://github.com/locuslab/tofu/blob/main/config/finetune.yaml), and [forgetting](https://github.com/locuslab/tofu/blob/main/config/forget.yaml) configs.

src/arcsf/config/config_class.py
philswatton: Have implemented but switched the property orders as `@classmethod` should be first

src/arcsf/models/model.py
philswatton: Yes, should be `False` by default (my mistake!)

src/arcsf/config/experiment_config.py
```diff
@@ -0,0 +1,164 @@
...
+        return obj
+    return [obj]
+
+
+def generate_experiment_configs(
+    top_config_name: str,
+) -> None:
+
+    # Read in yaml file
+    with open(f"configs/experiment/{top_config_name}.yaml") as f:
```
jack89roberts: would be preferable for this not to make assumptions about what directory the script has been called from (i.e. as it's a relative path this is assuming the experiment config is always created when running python from the parent `arc-selective-forgetting` directory)

scripts/train.py
```diff
@@ -0,0 +1,93 @@
...
+from arcsf.data.tofu import load_tofu
+from arcsf.utils import seed_everything
+
+
+def main(experiment_name):
+    start_time = datetime.strftime(datetime.now(), "%Y%m%d-%H%M%S-%f")
+
+    # Step 1: Process configs to dicts
+    experiment_config = ExperimentConfig.from_yaml(
+        f"configs/experiment/{experiment_name}.yaml"
```
jack89roberts: See other comment re making assumptions about what directory the script has been called from

jack89roberts (commented):
Played with this a bit this morning to submit a couple of longer training jobs.

- Added another gpt2/experiment config to run for 100 epochs (feel free to remove/tidy them up)
- Added `save_total_limit` as an arg to avoid saving a lot of checkpoints (reduce to 2 max, best and last), but this will likely need to be managed/reduced further. I did that after I realised the two jobs I have running will end up saving GPT2 200+ times...
- Remove a colon from the placeholder Q/A preprocessing (EDIT: doesn't look like this made a difference)
- Reduced walltime for the new longer example job only but you will likely want to do this for the original one too - it asks for 3 days which likely was why you were having to wait a long time to get on the queue.
- Stuck a datetime string in the temp save path to avoid overwriting the model you saved before

jack89roberts:
Leaving notes here as I go:

- `Trainer` output directory should be job specific to avoid jobs overwriting each other's stuff (but also maybe should be a temporary directory that we delete after, as we're not going to want to keep hundreds of checkpoints)
- Train script needs options to save models trained on the whole dataset (forget + retain), and only the retain split. The retain split model is the gold standard and the whole dataset model is the one that goes to the forgetting pipeline.

src/arcsf/config/experiment_config.py
```diff
@@ -0,0 +1,164 @@
...
+    def from_dict(cls, dict) -> "ModelConfig":
+        """Create a FineTuningConfig from a config dict.
+
+        Args:
+            config: Dict that must contain "model_id", "model_kwargs", and
+                "trainer_kwargs" keys.
+
+        Returns:
+            FineTuningConfig object.
+        """
```
jack89roberts: incorrect return type and docstring

jack89roberts:
I made a bit of a mess of commits, sorry. The last ones add a quick train on retain or everything workaround by adding a `train_type` to the experiment config (should be `retain` or `all`, maybe `forget` in future). Not sure it's the right place for it.

src/arcsf/models/config_class.py
philswatton: My inclination has been to have separate model and forgetting configs, with fine-tuning tied to the model config. The model config should then be re-used in forgetting (even if only to get the save destination path)

So in my very loose plan, forgetting config should contain the forgetting-relevant trainer class name and kwargs, or if re-using training kwargs some arg to handle that (maybe passed to the experiment config).

src/arcsf/models/config_class.py
philswatton: Have implemented

src/arcsf/models/trainer.py
philswatton: Have replaced as above by passing a dict of kwargs, which is processed earlier on in the model config

src/arcsf/models/config_class.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class ModelConfig(Config):
+
+    def __init__(
+        self,
+        model_id: str,
+        model_kwargs: dict,
+        trainer_kwargs: dict,
+        use_wandb: bool,
```
jack89roberts: Is it necessary to have WandB stuff in the model config (could it just be in the experiment config)?

jack89roberts:
I'm also looking at it on the forget branch but there are a few things that need updating after switching to just having a dict of training args, e.g. before `per_device_train_batch_size` was shortened to `train_batch_size` in the configs

src/arcsf/models/trainer.py
```diff
@@ -0,0 +1,54 @@
...
+    dataset: Dataset,
+    trainer_kwargs: dict,
+    use_wandb: bool,
+) -> Trainer:
+
+    # Load training arguments
+    training_args = TrainingArguments(
+        **trainer_kwargs,
+        overwrite_output_dir=True,
+        report_to="wandb" if use_wandb else None,
```
jack89roberts: This behaviour's going to change in the next major version of `transformers` it seems, but setting `report_to=None` actually translates to "report to everything" (it defaults to "all"). `report_to="none"` or `report_to=[]` should do what's intended here.

src/arcsf/models/config_class.py
philswatton: No - this is due to be removed after the previous changes and is WIP

philswatton: This is now done

src/arcsf/models/trainer.py
philswatton: Ah good spot, thank you

src/arcsf/config/experiment_config.py
philswatton: docstrings are TODO atm - this one's just a holdover from the LoCoMoSeT file I copied it from. will fix when I get around to doing docstrings

src/arcsf/models/config_class.py
philswatton: I've modified the `load_trainer` fn to take a `trainer_type` string w/ associated dict to manage that string. Can talk about about it once the PR is ready for review

philswatton: Have also pegged the data collator class selection to this

src/arcsf/config/experiment_config.py
philswatton: have switched to producing the absolute path

scripts/train.py
philswatton: as above

src/arcsf/__init__.py
philswatton: I slightly prefer easier imports for the functions being used in scripts (i.e. these ones), but it's not a strong preference so will remove

src/arcsf/config/config_class.py
philswatton: Have changed

src/arcsf/models/model.py
philswatton: Have made config dependent and `FALSE` by default

src/arcsf/config/experiment_config.py
philswatton: Now added

scripts/gpt2_longer_all.sh
```diff
@@ -0,0 +1,23 @@
...
+#SBATCH --gpus 1
+#SBATCH --cpus-per-gpu 36
+#SBATCH --output ./slurm_logs/example_experiment-%j.out
+
+# Load required modules here (pip etc.)
+module purge
+module load baskerville
+module load Miniconda3/4.10.3
+export CONDA_PKGS_DIRS=/tmp
+eval "$(${EBROOTMINICONDA3}/bin/conda shell.bash hook)"
```
jack89roberts, line 16: not a comment for this PR but note to self/selves to try `poetry` on Baskerville since MTQE have been using it

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+        # Load in other configsg
+        self.data_config = DataConfig.from_yaml(
+            os.path.join(arcsf.constants.DATA_CONFIG_DIR, f"{data_config}.yaml")
+        )
+        model_dir = os.path.join(arcsf.constants.MODEL_CONFIG_DIR, model_config)
+        self.model_config = ModelConfig.from_yaml(
+            os.path.join(model_dir, f"{model_config}.yaml"),
+            os.path.join(model_dir, "hyperparameters", f"{hyperparameter_config}.yaml"),
+        )
+
```
jack89roberts: Since the constants are `Path`s you should be able to do this (just a bit neater):


```suggestion
        # Load in other configs
        self.data_config = DataConfig.from_yaml(
            arcsf.constants.DATA_CONFIG_DIR / f"{data_config}.yaml"
        )
        model_dir = arcsf.constants.MODEL_CONFIG_DIR / model_config
        self.model_config = ModelConfig.from_yaml(
            model_dir / f"{model_config}.yaml",
            model_dir / "hyperparameters" / f"{hyperparameter_config}.yaml",
        )
```

scripts/train.py
```diff
@@ -0,0 +1,104 @@
...
+    parser = argparse.ArgumentParser(
+        description="""
+        This script takes some paths to dataset, and trainer configs, along with
+        arguments to the dataset config as inputs. The arguments to the dataset
+        config specify which dataset pair to perform the experiment on.
+
+        It then trains both models for the experiment with the parameters specified
+        in the trainer config, and logs the results to wandb. See README for more
+        information.
+        """
```
jack89roberts: Docstring looks out of date

scripts/train.py
```diff
@@ -0,0 +1,104 @@
...
+    dataset = FinetuneDataset(
+        data=dataset,
+        tokenizer=tokenizer,
+        qa_formatter=qa_formatter_basic,
+    )
+
+    # Step 7: Load trainer
+    experiment_config.model_config.trainer_kwargs["output_dir"] = (
+        f"{save_dir}/{experiment_config.model_config.trainer_kwargs['output_dir']}"
+    )
```
jack89roberts, line 59: can be left as is but wondering if this would be better done either in the model config class or `load_trainer` (just so it's less likely to be missed if using them elsewhere)

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+    """
+    if isinstance(obj, list):
+        return obj
+    return [obj]
+
+
+def _make_config_name(
+    top_config_name: str,
+    n: int,
+) -> str:
```
jack89roberts: Again minor but throughout this file short function names + args I'd prefer to be single line:
```suggestion
def _make_config_name(top_config_name: str, n: int) -> str:
```

If you remove the comma after the last arg `black` will auto-format them (or just re-add the comma if it wouldn't fit on a single line)

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+) -> dict:
+    # Generate combinations
+    combinations = _generate_combos_from_dict(combo_dict, full)
+
+    # Process model kwargs
+    for n, _ in enumerate(combinations):
+        combinations[n]["hyperparameter_config"] = combinations[n]["model_config"][1]
+        combinations[n]["model_config"] = combinations[n]["model_config"][0]
+        combinations[n] = {**combinations[n], **wandb_kwargs}
+
```
jack89roberts, line 83: Why there's `["model_config"][1]` and `["model_config"][0]` confused me. I see where it's coming from now but a comment here would be helpful (about how model + params are specified in the top level config), maybe also think if there's a better name for it in the configs e.g. `model_and_params` though that's pretty long and clunky.

```suggestion
    for n, comb in enumerate(combinations):
        # Top level model_config contains [model, hyperparameters] configs
        comb["model_config"] = comb["model_config"][0]
        comb["hyperparameter_config"] = comb["model_config"][1]
        combinations[n] = {**comb, **wandb_kwargs}

```

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+    sweep_dict_keys, sweep_dict_vals = zip(*combo_dict.items())
+    combinations = [
+        dict(zip(sweep_dict_keys, v)) for v in product(*list(sweep_dict_vals))
+    ]
+
+    # Initialise train_type
+    train_type = "retain"
+    if full:
+        train_type = "full"
+
```
jack89roberts: the use of `full` and `train_type` here (and in other functions) will likely need to change when forgetting is added in, but can do that in the other PR (Edit: And I've made another comment that might change this anyway)

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+        combinations[n]["model_config"] = combinations[n]["model_config"][0]
+        combinations[n] = {**combinations[n], **wandb_kwargs}
+
+    # Return
+    return combinations
+
+
+def generate_experiment_configs(
+    top_config_name: str,
+) -> None:
```
jack89roberts: Docstring to explain a bit about what's going on with the top level configs would be helpful here (also another function where the `def ... -> None` can all be on one line)

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+
+    # Read in yaml file
+    with open(
+        os.path.join(arcsf.constants.EXPERIMENT_CONFIG_DIR, f"{top_config_name}.yaml")
+    ) as f:
```
jack89roberts:
```suggestion
    with open(arcsf.constants.EXPERIMENT_CONFIG_DIR / f"{top_config_name}.yaml") as f:
```

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+            if (
+                full_combo["model_config"] == retain_combo["model_config"]
+                and full_combo["seed"] == retain_combo["seed"]
+                and full_combo["hyperparameter_config"]
+                == retain_combo["hyperparameter_config"]
+            ):
+                retain_combinations[n]["full_model_name"] = _make_config_name(
+                    top_config_name, m
+                )
+
```
jack89roberts: Not sure how much of an improvement it is in the end but was wondering whether some of this could be refactored along the lines of the below to avoid needing the big `if` at the end to recover the full model id (not tested + some of the other functions would need to be changed slightly):

```python
data_configs = top_config["combination"].pop("data_config")
combinations = generate_combos_from_dict(top_config["combination"], wandb_args):
retain_configs = []
full_configs = []
for full_idx, c in enumerate(combinations):
    # Model trained on whole dataset 
    fc = deepcopy(c)
    fc["data_config"] = top_config["full_data_config"]
    fc["train_type"] = "full"
    full_configs.append(fc)

    # Models trained on different retain splits
    for dc in data_configs:
        rc = deepcopy(c)
        rc["data_config"] = dc
        rc["full_model_name"] = _make_config_name(top_config_name, full_idx)
        rc["train_type"] = "retain"
        retain_configs.append(rc)
```

Ok to leave as is if this is missing something/not an improvement.

Also, things will get messier when adding forget models, I guess that would be an extra nested for loop...

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+    ) as f:
+        top_config = yaml.safe_load(f)
+
+    # Loop over, construct combo dict
+    combo_dict = {}
+    for key, value in top_config["combinations"].items():
+        combo_dict[key] = _listify(value)
+
+    # Construct same dict with full dataset
+    combo_dict_full = copy(combo_dict)
```
jack89roberts: `deepcopy` would be safer here (just `from copy import deepcopy`)

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+                and full_combo["seed"] == retain_combo["seed"]
+                and full_combo["hyperparameter_config"]
+                == retain_combo["hyperparameter_config"]
+            ):
+                retain_combinations[n]["full_model_name"] = _make_config_name(
+                    top_config_name, m
+                )
+
+    # All combinations to generate yaml files for
+    all_combinations = [*full_combinations, *retain_combinations]
```
jack89roberts:
```suggestion
    all_combinations = full_combinations + retain_combinations
```

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+
+    # Check this is a reasonably number of jobs for an array of training jobs
```
jack89roberts:
```suggestion
    # Check this is a reasonable number of jobs for an array of training jobs
```

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+        )
+        with open(file_name, "w") as f:
+            yaml.dump(combo, f)
+
+    # Check whether to generate baskerville scripts:
+    if top_config["use_bask"]:
+
+        # Create directory for train script for experiment if it doesn't exist
+        train_dir = os.path.join(arcsf.constants.PROJECT_DIR, "train_scripts")
+        if not os.path.exists(train_dir):
```
jack89roberts:
```suggestion
        # Create directory for train script for experiment if it doesn't exist
        train_dir = arcsf.constants.PROJECT_DIR / "train_scripts"
        if not train_dir.is_dir():
```

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+            os.mkdir(train_dir)
+
+        # Get jinja template
+        environment = Environment(
+            loader=FileSystemLoader(
+                os.path.join(arcsf.constants.PROJECT_DIR, "src", "arcsf", "config")
```
jack89roberts:
```suggestion
                arcsf.constants.PROJECT_DIR / "src" / "arcsf" / "config"
```

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+    def save(self, path: str) -> None:
+        """Save the config class into a yaml file.
+
+        Args:
+            path: Path and file name under which to save the config class.
+        """
+        with open(path, "w") as f:
+            yaml.dump(self.to_dict(), f)
+
+    def init_wandb(self, job_type: str) -> None:
```
jack89roberts, line 290: maybe default `job_type` to `self.train_type`?

src/arcsf/models/model.py
jack89roberts:
```suggestion
    add_token_to_model = False
```

src/arcsf/config/experiment.py
```diff
@@ -0,0 +1,304 @@
...
+            job_name=job_name,
+            walltime=top_config["bask"]["walltime"],
+            node_number=top_config["bask"]["node_number"],
+            gpu_number=top_config["bask"]["gpu_number"],
+            cpu_per_gpu=top_config["bask"]["cpu_per_gpu"],
+            array_number=len(all_combinations) - 1,
+            script_name="scripts/train.py",
+            experiment_file=f"{top_config_name}/experiment",
+        )
+        with open(os.path.join(train_dir, f"{job_name}.sh"), "w") as f:
```
jack89roberts:
```suggestion
        with open(train_dir / f"{job_name}.sh", "w") as f:
```

jack89roberts (changes requested):

jack89roberts:
^ mostly some small refactoring/docs comments. I haven't tried running it but looks good to me. Probably some upcoming headaches thinking about adding in forgetting nicely.

src/arcsf/constants.py
```diff
@@ -0,0 +1,21 @@
+from pathlib import Path
+
+PROJECT_DIR = Path(__file__, "..", "..", "..").resolve()
+CONFIG_DIR = Path(PROJECT_DIR, "configs")
+
+
+def _get_project_dir(location):
```
jack89roberts: maybe `_get_config_dir`?

src/arcsf/constants.py
```diff
@@ -0,0 +1,21 @@
...
+    """Helper function for creating project config paths.
+
+    Args:
+        location: Directory inside PROJECT_ROOT/configs/ to create path for
+
+    Returns:
+        String giving path of PROJECT_ROOT/configs/location
+    """
+    return str((CONFIG_DIR / location))
```
jack89roberts: may not be necessary to convert to a string here (but ok to leave if you prefer)

scripts/train.py
philswatton: Have updated

scripts/train.py
philswatton: As discussed, will leave for the forgetting PR

src/arcsf/config/experiment.py
philswatton: Have done

src/arcsf/config/experiment.py
philswatton: Have adopted your suggestion

src/arcsf/config/experiment.py
philswatton: Have added docstrings

src/arcsf/config/experiment.py
philswatton: Have replaced `copy` with `deepcopy`

src/arcsf/constants.py
philswatton: Have removed

src/arcsf/config/experiment.py
philswatton: I made this take `job_type` in anticipation of this being called during fine-tuning (`job_type = train`), forget (`job_type = forget`), and evaluation jobs (`job_type = eval`) - hence it isn't based on `self` but rather is an argument passed in the script calling it

src/arcsf/config/experiment.py
philswatton: Have used this refactor

src/arcsf/config/experiment.py
philswatton: Have implemented your other comment

src/arcsf/constants.py
philswatton: Have renamed

src/arcsf/config/experiment.py
jack89roberts: 👍

jack89roberts (approved):

PR #17, opened 2024-04-24
Dummy TOFU data

### LINKED ISSUE #16: Dummy TOFU dataset for tests

jack89roberts:
Adds a dummy TOFU-style dataset (and models) and switches the TOFU tests to use it.

Closes #16


tests/test_tofu.py
```diff
@@ -13,7 +20,7 @@ def _check_dropped_qs(granularity, stratify, forget_random, debug_dict):
     """check correct number of questions dropped per author/indices lie in range of
     forget authors -> not needed if random"""
 
-    forgot_q_frac = 0.1
+    forgot_q_frac = 0.34
```
J-Dymond: Is this necessary if defined globally? 
```suggestion
```

J-Dymond (approved):

tests/test_tofu.py
jack89roberts: ah yes I missed removing that one, thanks!

PR #18, opened 2024-04-25
9 forgetting pipeline

jack89roberts:
Adds a `Forgetter` class for Gradient Ascent, Gradient Difference, KL, and "I don't know forgetters", which all inherit from the HuggingFace `Trainer`, only modifying `compute_loss`.

Now updated to run with new top level config structure etc. I have successfully run jobs on Baskerville from this branch, but I have not fully verified the outputs are expected. There are runs on `wandb` under the `tofu-test` groups, e.g. https://wandb.ai/turing-arc/selective-forgetting/runs/dg1r098f/overview
 
### TODO as part of #25 or in another PR:

- Currently it's not possible to do forgetting with an `eval_dataset` / any kind of evaluation during training as the `evaluate` function in the trainer class doesn't know what to when given 2 data inputs (forget / retain). The [`evaluate`](https://github.com/alan-turing-institute/arc-selective-forgetting/blob/e2b8f87a937a24faf2f5a2ecca431a4660f33690/src/arcsf/forget/base.py) function in `arcsf.forget.base` need to be implemented to use Jack D's evaluation code (and the trainer should be initialised with `eval_dataset` set to whatever our appropriate eval dataset instance is).


src/arcsf/forget/base.py
```diff
@@ -0,0 +1,47 @@
...
+            return_outputs: Whether to return the outputs of the model or just the loss.
+
+        Returns:
+            The unlearning loss of the model, or the loss and the outputs of the model
+            if return_outputs is True.
+        """
+        raise NotImplementedError(
+            "A Forgetter child class implementing compute_loss should be used"
+        )
+
```
J-Dymond, line 39: Just some general comments on the evaluate method here. Does it need to take the dataset as an argument? In the `evaluate_model` function in the evaluation branch it takes the config which defines our datasets. 

It also takes: 
```
model: torch.nn.Module,
base_truth_ratios_path: str,
tokenizer: transformers.AutoTokenizer,
experiment_config: dict,
random_seed: int,
**generate_kwargs: dict,
```
So the only requirement beforehand is that the baseline retain model has run the script calculating its truth ratios (`all_eval.py`). Then we can point to that directory in the config? I had a look at the configs, can we define a `retain_model_path` when running forget experiments?  Everything else should be defined in the script running the experiment. 

Otherwise, I can adjust the `evaluate_model` code, such that it takes as an optional argument `eval_dataset`. This can be the output of load_dataset, which contains both splits of the data. Then the `evaluate_model` function will be able to parse these using the dictionary keys `'retain'` and `'forget'`.

src/arcsf/config/experiment.py
```diff
@@ -77,14 +66,62 @@ def generate_combos_from_dict(
...
     for n, _ in enumerate(combinations):
-        combinations[n]["hyperparameter_config"] = combinations[n]["model_config"][1]
-        combinations[n]["model_config"] = combinations[n]["model_config"][0]
         combinations[n] = {**combinations[n], **wandb_kwargs}
 
     # Return
     return combinations
 
 
+def make_config(
```
philswatton, line 75: Given the changes to configs, scripts/README.md will need to be updated to describe the new config structure + set out any limitations to what can be contained in a config + point to new examples rather than old ones

src/arcsf/forget/base.py
philswatton: Let's save solving this problem for #25, but if I've understood correctly this method overwrites `trainer.evaluate` so it would be difficult to change the arguments. I think Jack had in mind that we might want to run the evaluation in the same script as forgetting - reading your comment I'm not sure this will be possible

philswatton (approved):
LGTM, this has been run on Baskerville so will merge in

PR #25, opened 2024-05-13
10 add evaluation pipeline

### LINKED ISSUE #10: Add evaluation pipeline
Utility and forgetting evaluation as per TOFU

J-Dymond:
# Evaluation pipeline

Has a few utils files containing metrics and utility functions, and some scripts which perform evaluation on a selected model. Will briefly go over the evaluation scripts as they are at the moment, and the changes I made to the evaluation dataset class. The evaluation scripts can be run periodically throughout training to allow get a clearer picture of model performance as it is being trained. 

### `quantitative_eval.py`
Performs quantitative evaluation over a test set with a selected model, it compares ground truth inputs against *perturbed* inputs like in the paper. Unlike the paper we don't generate these, rather they are answers to different questions, randomly sampled from within the same author. In the future we can change this according to our work package. The script outputs the truth ratio values, and raw losses which are outputted as a numpy array for further processing. Currently if running as __main__ these are saved to a `.np` file in a separate folder in the parent folder of that where the model weights are stored.

### `qualitative_eval.py`
This performs a qualitative evaluation of the model. Loops over the test data and generates an output answer according to the input question. These are both printed along with the target printed along with the target to allow qualitative comparison against the target answer. 

### `EvalQADataset()` Changes
I made some changes to allow the quantitative evaluation script to work. Namely adding a batch formatter which when given a question, outputs input IDs, labels, and attention masks with appropriate padding for batch computation. Furthermore, a method which locates *perturbed* answers is added, which when given a question index will locate a random question pertaining the same author which can be used as a perturbed answer.


src/arcsf/data/tofu.py
```diff
@@ -154,4 +154,10 @@ def load_tofu(
         all_data[forget_indices]
     ), Dataset.from_dict(all_data[retain_indices])
 
+    setattr(forget_set, "TOFU_NUM_AUTHORS", TOFU_NUM_AUTHORS)
+    setattr(forget_set, "TOFU_Q_PER_AUTHOR", TOFU_Q_PER_AUTHOR)
+
+    setattr(retain_set, "TOFU_NUM_AUTHORS", TOFU_NUM_AUTHORS)
+    setattr(retain_set, "TOFU_Q_PER_AUTHOR", TOFU_Q_PER_AUTHOR)
```
jack89roberts: These can be: `forget_set.TOFU_NUM_AUTHORS = TOFU_NUM_AUTHORS`, and slight preference to rename to `forget_set.num_authors = TOFU_NUM_AUTHORS` so the names used in the rest of the code are less TOFU specific. Also, I've not looked at the rest of the code yet but are these the right values to be using? E.g. the forget/retain sets may have fewer authors/questions than the full dataset. If the values do need to be whole-dataset values then maybe make that clear in the names (which could mean ignoring the renaming suggestion above and keeping TOFU_...).

src/arcsf/data/data_module.py
```diff
@@ -84,9 +100,12 @@ def __init__(
         self,
         data: Dataset,
```
jack89roberts: Not a change in this PR so sorry I didn't pick up before, but is this actually meant to be a HuggingFace `Dataset` (`datasets.Dataset`) rather than a `torch` dataset (and throughout the code in typehints, docstrings etc.)? Unless we're explicitly making a torch dataset somewhere we're likely always working with HuggingFace datasets as that's what `load_dataset` returns.

src/arcsf/data/data_module.py
```diff
@@ -107,31 +129,119 @@ def __init__(
...
+
+        Returns:
+            formatted: batch formatted version of the perturbed inputs along with
+                formatted ground truth inputs
+        """
+        inp, _ = qa
+        author_q_n = row % self.data.TOFU_Q_PER_AUTHOR
+        perturbed_options = self.data[
+            row - author_q_n : row + self.data.TOFU_Q_PER_AUTHOR - author_q_n
+        ]["answer"]
```
jack89roberts: My understanding of this via an example:

```
row = 45
TOFU_Q_PER_AUTHOR = 20
author_q_n = 45 % 20 = 5
self.data[ 45 - 5 : 45 + 20 - 5] = self.data[40 : 60]
```
-> So is this trying to get all the answers for the relevant author? (then deleting the original answer on the next line)

But, as per previous comment, is `self.data` the whole dataset here or a (forget/retain) subset? If a subset will computing the indices this way work as expected? Could you use the `author_index` / `quetion_index` columns added in `load_tofu` instead, for example?

src/arcsf/data/data_module.py
```diff
@@ -107,31 +129,119 @@ def __init__(
...
+        del perturbed_options[author_q_n]
+        perturbed_options = perturbed_options[: self.n_perturbed]
```
jack89roberts: Should this be a random sample?

src/arcsf/data/data_module.py
```diff
@@ -107,31 +129,119 @@ def __init__(
...
+        formatted = [None] * (self.n_perturbed + 1)
+        formatted[0] = self.batch_formatter(qa)
+        for p_idx, perturbed in enumerate(perturbed_options):
+            formatted[p_idx + 1] = self.batch_formatter((inp, perturbed))
+        return formatted
```
jack89roberts:
```suggestion
        formatted = [self.batch_formatter((inp, perturbed)) for perturbed in perturbed_options]
        return [self.batch_formatter(qa)] + formatted
```

src/arcsf/data/data_module.py
```diff
@@ -107,31 +129,119 @@ def __init__(
...
+        self.device = device
 
         if loss_type == "idk":
             self.idk = get_idk_responses()
             self.answer_sampler = self.get_idk
         else:
             self.answer_sampler = self.get_answer
 
+        if qualitative_eval:
+            self.format = self.qualitative_formatter
```
jack89roberts: Placeholder for further thought/discussion - would we like it to be possible to do both qualitative and quantative from a single EvalQADataset instance?

src/arcsf/data/data_module.py
```diff
@@ -107,31 +129,119 @@ def __init__(
...
+    def qualitative_formatter(self, qa, _):
+        encoded_inp = self.tokenizer(
+            qa[0], return_tensors="pt", add_special_tokens=True
+        )
+        encoded_tar = self.tokenizer(
+            qa[1], return_tensors="pt", add_special_tokens=True
+        )
+        return (encoded_inp, encoded_tar)
+
+    def batch_formatter(
```
jack89roberts: maybe easier to discuss separately but in the HuggingFace trainer the batching is usually handled by a separate data collator - there's an example on the forgetting branch. So this maybe could be replaced with a function from elsewhere.

src/arcsf/data/data_module.py
```diff
@@ -107,31 +129,119 @@ def __init__(
...
 
     def __getitem__(self, idx):
-        input = self.data[idx]["question"]
-
-        target = self.answer_sampler(self.data[idx]["answer"])
+        inp = self.data[idx]["question"]
+        tar = self.answer_sampler(idx)
 
-        return self.tokenizer(input), self.tokenizer(target)
+        return self.format((inp, tar), idx)
```
jack89roberts: Placeholder whilst I look at the code: Is it necessary to return `idx` here?

src/arcsf/eval/metrics.py
```diff
@@ -0,0 +1,39 @@
+import torch
```
jack89roberts, line 1: Some comments/type hints/docstrings/references back to the TOFU paper would be nice here to help understand what's going on - mainly for truth ratio.

src/arcsf/eval/qualitative_eval.py
```diff
@@ -0,0 +1,109 @@
...
+    Args:
+        model : Transformers model used to perform evaluation on
+        tokenizer : Tokenizer for the purpose of decoding model output
+        dataset : Dataset to perform evaluation on
+        n_inputs : number of inputs to sample
+        random_seed : Random seed for the dataloader
+        generate_kwargs : keyword arguments for the model.generate() method
+
+    Returns:
+        None
```
jack89roberts:
```suggestion
```

Not needed as it doesn't add anything over the `-> None` type hint.

src/arcsf/eval/qualitative_eval.py
```diff
@@ -0,0 +1,109 @@
...
+def qualitative_eval(
+    model: torch.nn.Module,
+    tokenizer: AutoTokenizer,
+    dataset: Dataset,
+    n_inputs: int,
+    random_seed: int,
+    **generate_kwargs: dict,
+) -> None:
+    """Performs qualitative evaluation of the selected model over selected data.
+    Prints the generated text give a question followed by the ground truth target.
```
jack89roberts:
```suggestion
    Prints the generated text given a question followed by the ground truth target.
```

src/arcsf/eval/qualitative_eval.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description=(
+            "Runs qualitative evaluation, comparing outputs of model"
+            " against target strings."
+        )
+    )
+    parser.add_argument("directory", type=str, help="Relative path to model directory.")
```
jack89roberts: `directory` -> `model_path`?

src/arcsf/eval/qualitative_eval.py
```diff
@@ -0,0 +1,109 @@
...
+    # these are hardcoded for now
+    rand = 42
+    model = GPT2LMHeadModel.from_pretrained(model_dir)
+    tokenizer = GPT2Tokenizer.from_pretrained(model_dir)
+    model.config.pad_token_id = tokenizer.eos_token_id
+
+    experiment_config = yaml.safe_load(open(model_dir + "/experiment_config.yaml"))
+    forget_data, retain_data = get_data(
+        "tofu", random_seed=rand, **experiment_config["data_config"]
+    )
```
jack89roberts: If using the experiment config the path should be configurable (and it probably could be possible to get the model path from there too). Or could take a data config only (also with a configurable path).

src/arcsf/eval/quantitative_eval.py
```diff
@@ -0,0 +1,126 @@
...
+    data_loader = DataLoader(dataset, batch_size=batch_size, shuffle=False)
+    all_losses = torch.zeros((dataset.__len__(), n_perturbed + 1))
+    # loop over batches
+    for batch_idx, batch in enumerate(tqdm(data_loader, desc="Batch")):
+        gt_batch = batch[0]
+        batch_start_index = batch_idx * batch_size
+        batch_end_index = batch_idx * batch_size + len(gt_batch["input_ids"])
+        # don't need gradient
+        with torch.no_grad():
+            gt_dummy_model_output = model(**gt_batch)
```
jack89roberts: `gt_dummy_model_output` -> `gt_outputs`?

src/arcsf/eval/quantitative_eval.py
```diff
@@ -0,0 +1,126 @@
...
+        # get ground truth loss
+        gt_loss = get_loss(gt_dummy_model_output.logits, gt_batch["labels"])
+        all_losses[batch_start_index:batch_end_index, 0] = gt_loss
+        # loop over perturbed samples to get their losses
+        for perturbed_index in range(1, n_perturbed + 1):
+            p_batch = batch[perturbed_index]
+            with torch.no_grad():
+                p_dummy_model_output = model(**p_batch)
```
jack89roberts: `p_dummy_model_output` -> `p_outputs`?

src/arcsf/eval/quantitative_eval.py
```diff
@@ -0,0 +1,126 @@
...
+
+from arcsf.data.data_module import EvalQADataset, get_data, qa_formatter_autoregression
+from arcsf.eval.metrics import truth_ratio
+from arcsf.eval.utils import get_loss
+from arcsf.utils import get_device
+
+
+def quantitative_eval(
+    model: torch.nn.Module,
+    dataset: Dataset,
```
jack89roberts: is this expected to be an `EvalQADataset`?

src/arcsf/eval/quantitative_eval.py
```diff
@@ -0,0 +1,126 @@
...
+                p_dummy_model_output = model(**p_batch)
+            all_losses[batch_start_index:batch_end_index, perturbed_index] = get_loss(
+                p_dummy_model_output.logits, p_batch["labels"]
+            )
+    # calculate truth_ratio and return them along with losses
+    means = torch.mean(all_losses, dim=0)
+    print(f"Mean losses:\n\tground_truth:{means[0]}\n\tperturbed:{means[1:]}")
+    tr = truth_ratio(all_losses)
+
+    return tr, all_losses
```
jack89roberts: Does/should this return all the metrics (e.g. rouge etc.)?

tests/test_data_module.py
```diff
@@ -107,10 +107,7 @@ def test_idk_targets(data):
     # load idk type dataset - don't pass tokenizer or qa_formatter so we can look
     # directly at output.
     idk_set = EvalQADataset(
-        data,
-        _identity,
-        _identity,
-        loss_type="idk",
+        data, _identity, _identity, loss_type="idk", qualitative_eval=True
```
jack89roberts: It seems a bit odd to me that this requires `qualitative_eval` to be `True` to work

tests/test_eval.py
```diff
@@ -0,0 +1,168 @@
...
+    # argmax predictions for test targets
+    test_targets = test_outputs_correct.argmax(dim=1)
+    # flip random predictions for incorrect predictions
+    test_outputs_incorrect = test_outputs_correct * -1
+
+    correct = eval_accuracy(test_outputs_correct, test_targets)
+    assert correct["eval_accuracy"] == pytest.approx(1.0)
+
+    incorrect = eval_accuracy(test_outputs_incorrect, test_targets)
+    assert incorrect["eval_accuracy"] == pytest.approx(0.0)
```
jack89roberts: These should be exact (try to only use `pytest.approx` where it's definitely needed):

```suggestion
    assert correct["eval_accuracy"] == 1.0

    incorrect = eval_accuracy(test_outputs_incorrect, test_targets)
    assert incorrect["eval_accuracy"] == 0.0
```

And a test of an intermediate value (between 0 and 1) would be a nice to have.

tests/test_eval.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def test_conditional_probability():
```
jack89roberts, line 77: Same comments as accuracy test apply here

tests/test_eval.py
```diff
@@ -0,0 +1,168 @@
...
+    incorrect_losses[:, 1] = 0
+
+    eval_prob = conditional_probability(correct_losses)
+    assert torch.mean(eval_prob["conditional_probs"][0]).item() == pytest.approx(1.0)
+
+    eval_prob = conditional_probability(incorrect_losses)
+    assert torch.mean(eval_prob["conditional_probs"][0]).item() == pytest.approx(0.0)
+
+
+def test_ks_test():
```
jack89roberts, line 95: test of a non-zero value would be good too

tests/test_eval.py
```diff
@@ -0,0 +1,168 @@
...
+    labels = input_ids.clone()
+
+    test_output = model(
+        input_ids=input_ids, attention_mask=attention_mask, labels=labels
+    )
+
+    loss = get_loss(test_output.logits, labels)
+
+    assert isinstance(loss, torch.Tensor)
+    assert loss.grad_fn
```
jack89roberts: is this the right test for this? I don't think we care about gradients for the eval?

jack89roberts (changes requested):

src/arcsf/eval/quantitative_eval.py
J-Dymond: Yes

src/arcsf/eval/quantitative_eval.py
J-Dymond: Yes

src/arcsf/eval/quantitative_eval.py
J-Dymond: yes

src/arcsf/eval/quantitative_eval.py
J-Dymond: Can update this to also return the probability, the `all_losses` object should allow us to calculate this elsewhere, but will implement here too. I will also investigate where Rouge score should be returned.

src/arcsf/eval/qualitative_eval.py
J-Dymond: yes

src/arcsf/eval/qualitative_eval.py
J-Dymond: Yes, I agree. I can make this change to the argparser

src/arcsf/data/data_module.py
J-Dymond: I tried this and it broke a lot of tests, maybe we can have a look at this in a co-working session?

src/arcsf/data/data_module.py
J-Dymond: Yeah I could see that, I was keen to have as little logic as possible in the forward method to make it more efficient.. But I understand the bottleneck is generally the model inference. 

Maybe we could have an attribute which we can change to dictate the __getitem__ method? ie dataset.return_qualitative == True and dataset.return_quantitative == False, then when the __getitem__ method will look at these to work out what to return?

src/arcsf/data/data_module.py
J-Dymond: Perhaps we can discuss this in a co-working session?

src/arcsf/data/data_module.py
J-Dymond: Exactly, that's what its doing. But I recognise now this doesn't generalise to a shuffled set. As you say, it would be much better to use the author index to find the relevant items. I will make this change.

src/arcsf/data/data_module.py
J-Dymond: We could randomly sample, this was a placeholder that I forgot to develop further. Perhaps I could randomly shuffle the perturbed_options array to keep this simple?

src/arcsf/data/data_module.py
J-Dymond: applied, thank you!

src/arcsf/data/data_module.py
J-Dymond: it returns the output of self.format(), index is sometimes used in this function to identify the perturbed options

tests/test_data_module.py
J-Dymond: otherwise it returns batch formatted inputs, qualitative evaluation outputs a (question,answer) tuple

J-Dymond: By passing in the _identity function for the tokenizer and formatter it will just return the raw strings. Meaning we can directly compare them with the file they're sampled from. It just saves us having to decode the dataset output.

src/arcsf/data/data_module.py
J-Dymond: I think I've addressed this now

J-Dymond: Though, my new implementation runs self.data.filter() everytime __getitem__ is called. I will think of a way to calculate these at initialisation and store them?

src/arcsf/eval/qualitative_eval.py
J-Dymond: done thank you

src/arcsf/eval/qualitative_eval.py
J-Dymond: yes, this was a typo thank you

src/arcsf/data/data_module.py
J-Dymond: I've added a random shuffle now when its sampled, with the seed passed to the dataset

src/arcsf/data/data_module.py
J-Dymond: clarified type-hinting for these

src/arcsf/eval/quantitative_eval.py
J-Dymond: Revisiting the paper, the probability is used when there are multiple choice answers. That is, the real author/world fact questions. So here we're not using them. Though, perhaps this is something that needs addressing?

tests/test_eval.py
J-Dymond: Good point, I've removed it

tests/test_eval.py
J-Dymond: Done

src/arcsf/eval/quantitative_eval.py
J-Dymond: Rouge score will be better placed in qualitative eval, since the inputs are strings. Though maybe it would be better to get all of the metrics from a single script? I have issues here though, formatting the inputs when using batches.

src/arcsf/data/data_module.py
J-Dymond: the dataset now has the option of doing both

src/arcsf/eval/metrics.py
J-Dymond: will do

src/arcsf/data/data_module.py
J-Dymond: I implemented something like written above. Setting this up for batch_computation could be investigated in the future.

src/arcsf/data/data_module.py
J-Dymond: Discussed in co-working session. Can implement this when it is needed.

src/arcsf/data/tofu.py
J-Dymond: added this change

tests/test_eval.py
J-Dymond: added an intermediate test as well

jack89roberts:
- Ensure we understand what's going on in the eval (e.g. document shapes etc.)
  - including NaNs appearing in places 
- Compute and return `alternative=greater` in KS test (as well as default).
- Try to understand why IDK model comes out worse for forgetting

jack89roberts:
It would be nice if the tests capture some of what we were trying to think through earlier today, e.g. check truth ratio of one of the dummy forget models is larger than the dummy fine-tuned model and similar. I haven't checked back through the tests so it might be that you've already done that.

J-Dymond:
I've made some changes to my pull request now, I've added a function in `evaluate_model.py` called `evaluate_model` :

```
def evaluate_model(
    model: torch.nn.Module,
    base_truth_ratios_path: str,
    tokenizer: transformers.AutoTokenizer,
    experiment_config: dict,
) -> dict[float, float, float, float, float]:
```

It outputs a dictionary containing:

```
result_dict = {
        "mean_tr_retain": retain_tr,
        "mean_rouge_score": rouge_score,
        "forget_quality_1": forget_quality_one_sided,
        "forget_quality_2": forget_quality_two_sided,
        "model_utility": model_utilty,
    }
```
This should be everything we want to track in the wandb. I've added some tests testing it, and I've moved the old scripts I wrote into a `/scripts` folder within `eval`. The functions for these scripts are all contained in utils now. So within the `eval` folder there are just three files:

- `metrics.py` : containing the functions for the metrics
- `utils.py` : containing all of the functions used in the scripts folder and the evaluate_model function
- `evaluate_model.py` : containing the function for evaluating model, in hindsight, maybe this can be moved to utils.. but I leave that up to what you guys think is best

J-Dymond:
Just something minor I didn't explicitly point out in the above: the path for the base model truth ratios should currently be the relative path to where the **forget** truth ratios are stored. 

The all_eval script will calculate and save these, provided you give it the forget dataset. 

These are the only values that need to be stored locally for evaluate_model to run, everything else should be calculated within the function.

src/arcsf/data/data_module.py
```diff
@@ -69,11 +70,27 @@ def qa_formatter_basic(qa: tuple[str, str]) -> str:
     Returns:
         full_text: formatted question--answer pair
     """
-    question, answer = qa
+    question, answer = qa[0], qa[1]
     return f"Question: {question}\nAnswer: {answer}"
 
 
-class EvalQADataset(Dataset):
+def qa_formatter_autoregression(qa: tuple[str, str, int]) -> str:
```
jack89roberts: Maybe `qa_formatter_blank` or similar (i.e. nothing added other than question and answer).

src/arcsf/data/data_module.py
```diff
@@ -57,7 +58,7 @@ def get_data(
     return data
 
 
-def qa_formatter_basic(qa: tuple[str, str]) -> str:
+def qa_formatter_basic(qa: tuple[str, str, int]) -> str:
```
jack89roberts: Note to self: Forgetting branch has refactored QA formatters.

src/arcsf/data/data_module.py
```diff
@@ -107,34 +132,144 @@ def __init__(
...
+        encoded_inp = self.tokenizer(
+            qa[0], return_tensors="pt", add_special_tokens=True
+        )
+        encoded_tar = self.tokenizer(
+            qa[1], return_tensors="pt", add_special_tokens=True
+        )
+        return (encoded_inp, encoded_tar)
+
+    def batch_formatter(
+        self,
```
jack89roberts: Note to self: Data collators/padding choices.

src/arcsf/data/data_module.py
```diff
@@ -107,34 +132,144 @@ def __init__(
...
+        """
+        inp, _ = qa
+        author_n = self.data[row]["author_index"]
+        question_n = self.data[row]["question_index"]
+
+        with hf_progress_bars_disabled():
+            perturbed_options = self.data.filter(
+                lambda sample: sample["author_index"] == author_n
+                and sample["question_index"] != question_n
+            ).shuffle(seed=self.rand_gen.seed())
```
jack89roberts: I've not gone through to check but is setting the seed like this here ok or will it be giving the same perturbed samples every time? i.e. is this function used multiple times during an evaluation run and should/does it give different perturbed rows each time?

src/arcsf/data/tofu.py
```diff
@@ -154,4 +154,10 @@ def load_tofu(
         all_data[forget_indices]
     ), Dataset.from_dict(all_data[retain_indices])
 
+    forget_set.num_authors = TOFU_NUM_AUTHORS
+    forget_set.q_per_author = TOFU_Q_PER_AUTHOR
+
+    retain_set.num_authors = TOFU_NUM_AUTHORS
+    retain_set.q_per_author = TOFU_Q_PER_AUTHOR
```
jack89roberts: Are these still needed/used?

src/arcsf/eval/scripts/all_eval.py
```diff
@@ -0,0 +1,81 @@
...
+import yaml
+from transformers import GPT2LMHeadModel, GPT2Tokenizer
+
+from arcsf.data.data_module import EvalQADataset, get_data, qa_formatter_autoregression
+from arcsf.eval.utils import all_eval
+from arcsf.utils import get_device
+
+if __name__ == "__main__":
+    # currently this is our random seed
+    rand = 42
```
jack89roberts: if leaving a seed fixed here we should make an issue about it - I think it's only needed due to the way we're mimicing paraphrases/perturbed answers currently, so maybe it can be added to an issue about that.

src/arcsf/eval/scripts/all_eval.py
```diff
@@ -0,0 +1,81 @@
...
+
+    parser = argparse.ArgumentParser(
+        description=(
+            "Runs quantitative evaluation, comparing output logits of model against"
+            " targets. It currently saves the truth ratio and input-wise losses."
+        )
+    )
+    batch_size = 1
+    parser.add_argument("directory", type=str, help="Relative path to model directory.")
```
jack89roberts: `directory` -> `model_path`, `model_dir` or similar

src/arcsf/eval/scripts/analysis.py
```diff
@@ -0,0 +1,100 @@
...
+from arcsf.eval.utils import get_analysis_values, get_metrics, plot_cdf, plot_scatter
+
+
+def run_analysis(
+    base_model_path: str,
+    test_model_path: str,
+    verbose: bool = False,
+    plotting: bool = False,
+) -> dict[str, float, float, float, float]:
+    """
```
jack89roberts: Type hinting for dicts is just key type then value type (not one for each value)


```suggestion
) -> dict[str, float]:
    """
```

src/arcsf/eval/scripts/analysis.py
```diff
@@ -0,0 +1,100 @@
...
+        try:
+            results = run_analysis(base_dir, "temp/" + dir, plotting=True, verbose=True)
+        except FileNotFoundError:
+            continue
+        all_results[results.pop("experiment_name")] = results
+
+    with open("temp/results/all_results.json", "w") as f:
+        json.dump(all_results, f)
+
+    plt.savefig("temp/results/all_models_cdf.pdf")
```
jack89roberts: all directories etc. here should be configuarable ideally

src/arcsf/eval/scripts/qualitative_eval.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    args = parser.parse_args()
+    model_dir = args.model_path
+
+    # these are hardcoded for now
+    rand = 42
+    model = GPT2LMHeadModel.from_pretrained(model_dir)
+    tokenizer = GPT2Tokenizer.from_pretrained(model_dir)
+    model.config.pad_token_id = tokenizer.eos_token_id
```
jack89roberts: if leaving hardcoded for this PR make an issue about fixing this later

src/arcsf/eval/evaluate_model.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def evaluate_model(
+    model: torch.nn.Module,
+    base_truth_ratios_path: str,
+    tokenizer: transformers.AutoTokenizer,
+    experiment_config: dict,
+    **generate_kwargs: dict,
+) -> dict[float, float, float, float, float]:
+    """
```
jack89roberts:
```suggestion
) -> dict[str, float]:
    """
```

src/arcsf/eval/evaluate_model.py
```diff
@@ -0,0 +1,124 @@
...
+    config and the path to the baseline model truth ratios.
+
+    Args:
+        model: Model to perform the evaluation on
+        base_truth_ratios_path: path for the baseline mode truth ratios
+        tokenizer: tokenizer being used for the model
+        experiment_config: experiment config (currently a dictionary)
+
+    Returns:
+        dictionary : aggregated metrics for evaluation (generated using `get_metrics`)
```
jack89roberts:
```suggestion
        dictionary of aggregated metrics for evaluation (generated using `get_metrics`)
```

src/arcsf/eval/evaluate_model.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    # get datasets
+    forget_data, retain_data = get_data(
+        "tofu", random_seed=RAND, **experiment_config["data_config"]
+    )
+    # get available device
+    device = get_device()
+
+    # create datasets
+    RetainDataset = EvalQADataset(
```
jack89roberts: naming convention: `retain_dataset`

src/arcsf/eval/evaluate_model.py
```diff
@@ -0,0 +1,124 @@
...
+        retain_data,
+        tokenizer,
+        qa_formatter_autoregression,
+        "standard",
+        quantitative_eval=True,
+        qualitative_eval=True,
+        device=device,
+        random_seed=RAND,
+    )
+    ForgetDataset = EvalQADataset(
```
jack89roberts: naming convention: `forget_dataset`

src/arcsf/eval/utils.py
```diff
@@ -0,0 +1,299 @@
...
+    return losses
+
+
+def ecdf(x):
+    xs, _ = torch.sort(x)
+    ys = torch.arange(1, len(xs) + 1) / float(len(xs))
+    return xs, ys
+
+
+def combine_dicts(forget_dict, retain_dict):
```
jack89roberts: could have a more informative name than `combine_dicts`

src/arcsf/eval/metrics.py
```diff
@@ -0,0 +1,98 @@
...
+         accuracy : accuracy metric in the form of a dictionary
+    """
+    preds = logits.argmax(dim=-1)
+    accuracy = torch.mean(torch.eq(preds, labels).float())
+    return {"eval_accuracy": accuracy}
+
+
+def conditional_probability(
+    token_normalised_losses: torch.Tensor,
+) -> dict[torch.Tensor]:
```
jack89roberts:
```suggestion
) -> dict[str, torch.Tensor]:
```

but see next comment

src/arcsf/eval/metrics.py
```diff
@@ -0,0 +1,98 @@
...
+            shape: n_samples x n_perturbed
+
+    Returns:
+        conditional_probs: conditional probabilities in the form of a dictionary
+    """
+    probs = torch.exp(-1 * token_normalised_losses)  # shape : n_samples x n_perturbed
+    sum = torch.sum(probs, dim=-1)  # shape : n_samples
+    # transpose here ensures that the n_samples dimension is in the right place
+    cond_probs = probs.T / sum  # shape : n_perturbed x n_samples
+    return {"conditional_probs": cond_probs}
```
jack89roberts: why not just return `cond_probs` here? The only place I can see that this function is used the values are immediately extracted from the dictionary again anyway.

src/arcsf/eval/metrics.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def eval_rouge_recall(gen_output: str, ground_truth: str) -> dict[float, float]:
```
jack89roberts:
```suggestion
def eval_rouge_recall(gen_output: str, ground_truth: str) -> dict[str, float]:
```

src/arcsf/eval/metrics.py
```diff
@@ -0,0 +1,98 @@
...
+    Abstraction of the RogueScorer class call, calculates the ROUGE score of the
+    generated output. rougeL and rouge1 are outputted but the TOFU codebase only used
+    rougeL.
+
+    Args:
+        gen_output : generated output
+        ground_truth : ground truth string
+
+    Returns:
+        dict[float, float]: scores
```
jack89roberts:
```suggestion
        ROUGE scores
```

src/arcsf/eval/metrics.py
```diff
@@ -0,0 +1,98 @@
...
+        normalised_losses : Token length normalised losses.
+            shape: n_samples x number_perturbed
+
+    Returns:
+        truth_ratio : the truth ratio
+    """
+    # getting conditional probability of the losses, this does perform a transpose
+    # which doesn't affect the output values, but also isn't strictly necessary.
+
+    # cond_probs shape: n_perturbed x n_samples
```
jack89roberts: is this technically `n_perturbed + 1` as it includes the correct answer as well?

src/arcsf/eval/utils.py
```diff
@@ -0,0 +1,299 @@
...
+
+    Returns:
+        _description_
+    """
+    output_logits = output_logits[
+        ..., :-1, :
+    ].contiguous()  # shape: batch_size x (sequence_length-1) x vocab_size
+    shifted_labels = labels[
+        ..., 1:
+    ].contiguous()  # shape : batch_size x (sequence_length - 1)
```
jack89roberts:
```suggestion
    # shape: batch_size x (sequence_length-1) x vocab_size
    output_logits = output_logits[..., :-1, :].contiguous()

    # shape : batch_size x (sequence_length - 1)
    shifted_labels = labels[..., 1:].contiguous()  
```

src/arcsf/eval/utils.py
```diff
@@ -0,0 +1,299 @@
...
+    # output_logits.transpose(-1, -2) shape: batch_size x vocab x (sequence_length - 1)
+    loss = _loss_function(output_logits.transpose(-1, -2), shifted_labels).sum(
+        dim=-1
+    )  # loss shape: batch_size
```
jack89roberts:
```suggestion
    # loss shape: batch_size
    loss = _loss_function(output_logits.transpose(-1, -2), shifted_labels).sum(dim=-1)  
```

src/arcsf/eval/utils.py
```diff
@@ -0,0 +1,299 @@
...
+        "forget_tr": forget_dict["truth_ratios"],
+        "rouge_scores": retain_dict["rougeL_recall"],
+    }
+
+
+def get_metrics(
+    base_truth_ratios: torch.Tensor,
+    test_values: dict[torch.Tensor],
+) -> dict[float, float, float, float, float]:
+    """
```
jack89roberts:
```suggestion
) -> dict[str, float]:
    """
```

src/arcsf/eval/utils.py
```diff
@@ -0,0 +1,299 @@
...
+        "forget_quality_1": forget_quality_one_sided.item(),
+        "forget_quality_2": forget_quality_two_sided.item(),
+        "model_utility": model_utilty.item(),
+    }
+
+    return result_dict
+
+
+def all_eval(
+    model: torch.nn.Module,
```
jack89roberts: All model typehints maybe should be `transformers.PreTrainedModel` (but could be they all work with `torch` modules anyway)

```suggestion
    model: transformers.PreTrainedModel,
```

src/arcsf/eval/utils.py
```diff
@@ -0,0 +1,299 @@
...
+    # move model to device create dataloader, initialise all_losses tensor
+    model = model.to(device)
+    n_perturbed = dataset.n_perturbed
+    data_loader = DataLoader(dataset, batch_size=batch_size, shuffle=False)
+    output_dict = {
+        "all_losses": torch.zeros(
+            (dataset.__len__(), n_perturbed + 1), dtype=torch.float64
+        ),
+        "truth_ratios": torch.zeros(dataset.__len__()),
+        "rougeL_recall": torch.zeros(dataset.__len__()),
```
jack89roberts: Use `len` rather than calling `__len__` directly


```suggestion
        "all_losses": torch.zeros(
            (len(dataset), n_perturbed + 1), dtype=torch.float64
        ),
        "truth_ratios": torch.zeros(len(dataset)),
        "rougeL_recall": torch.zeros(len(dataset)),
```

src/arcsf/eval/utils.py
```diff
@@ -0,0 +1,299 @@
...
+        print(f"Target: {target}")
+
+        if batch_idx >= n_inputs:
+            break
+
+
+def get_analysis_values(
+    model_dir: str,
+) -> dict[np.ndarray, np.ndarray, np.ndarray, torch.Tensor, torch.Tensor]:
+    """
```
jack89roberts:
```suggestion
) -> dict[str, np.ndarray | torch.Tensor]:
```

src/arcsf/eval/utils.py
```diff
@@ -0,0 +1,299 @@
...
+
+    Returns:
+        vals : dictionary containing the values to be used in analysis
+    """
+    vals = dict()
+    # losses are of shape (n_samples x n_perturbed)
+    vals["forget_losses"] = np.loadtxt(model_dir + "/eval/forget/all_losses.txt")
+    vals["retain_losses"] = np.loadtxt(model_dir + "/eval/retain/all_losses.txt")
+    vals["rouge_scores"] = np.loadtxt(model_dir + "/eval/retain/rougeL_scores.txt")
+    # we re-calculate the truth ratio, since torch calculated many as NaNs
```
jack89roberts: could we check for nans and log a warning somewhere if they appear?

jack89roberts (changes requested):
I haven't got my head around it fully yet, will look again next week. I left a few comments about things being hardcoded but not for everything - for this PR they can be kept hardcoded but if so we should make an issue with everything that's left outstanding/will need to be changed for future runs/experiments.

src/arcsf/eval/utils.py
J-Dymond: I actually thought I set all of these to torch.Tensors. I'll have a look at that, but will add this change

src/arcsf/eval/utils.py
J-Dymond: I'm not sure how to log warnings, but will investigate

J-Dymond: Added this now, it checks all of them at the end, and logs an error if there is.

src/arcsf/eval/utils.py
J-Dymond: Made this change. Since I was using it three times I just created a variable. Is there a reason we prefer len() over calling __len__ ?

src/arcsf/data/data_module.py
J-Dymond: Made this change

src/arcsf/data/data_module.py
J-Dymond: It will give the same question indices for each question. We can remove the seed if this isn't the functionality we want?

src/arcsf/eval/evaluate_model.py
J-Dymond: Done

src/arcsf/eval/evaluate_model.py
J-Dymond: Done

src/arcsf/eval/scripts/qualitative_eval.py
J-Dymond: Random seed was the only hardcoded value, it is now given as an input argument, and picks a random value if it is not specified.

src/arcsf/data/tofu.py
J-Dymond: I don't think so, will remove.

J-Dymond: This was originally used when we sampled questions, but now we use the author index instead. So this isn't used any more.

src/arcsf/eval/metrics.py
J-Dymond: I kept this since the original TOFU code did it, I wasn't sure why they used a dictionary either but I thought there might have been some logging tool using it or something. 

Anyway.. I can change it since it isn't needed.

src/arcsf/eval/metrics.py
J-Dymond: addressed this

src/arcsf/eval/metrics.py
J-Dymond: Thank you, I think I missed this auto fill from my docstring plugin

src/arcsf/eval/metrics.py
J-Dymond: yes, this was a mistake, thank you for noticing

src/arcsf/eval/scripts/all_eval.py
J-Dymond: I have made it an input argument

src/arcsf/eval/scripts/analysis.py
J-Dymond: Ah yeah, sorry this was something I only intended to run locally to get our results.

src/arcsf/eval/utils.py
J-Dymond: Addressed this now with a docstring

src/arcsf/eval/scripts/analysis.py
J-Dymond: I've ammended this now. However, there is no longer a script to aggregate them all into a single plot. I can make something new, or we can do this as and when we need it.

jack89roberts:
I've rebased this on the forgetting pipeline branch - so merge the forget branch PR before this one. I'm going to spend some time this afternoon thinking about how to get the eval working during forgetting.

jack89roberts:
@J-Dymond (when you're back) - I'm struggling to understand the use of the quantitative/qualitative options in `EvalQADataset`. E.g. the `all_eval` script doesn't seem to do anything qualitative but it uses an `EvalQADataset` with qualitative set to true and things break if I switch it to false.

src/arcsf/eval/evaluate_model.py
```diff
@@ -0,0 +1,126 @@
...
+        forget_dataset,
+        1,
+        device,
+        tokenizer,
+        **generate_kwargs,
+    )
+
+    # combine dictionaries and load the baseline model truth ratios
+    analysis_vals = combine_dicts(forget_values, retain_values)
+    base_truth_ratios = torch.tensor(np.loadtxt(base_truth_ratios_path))
```
jack89roberts: I've not checked the paper as of writing this, but shouldn't this be using base truth ratios from the retain data and the forget data here (rather than just one set of base truth ratios that's used in both the retain and forget evals)?

jack89roberts:
Couple of observations:

- `all_eval` is very slow - can it be made faster? Maybe one for a different PR/issue.
- `all_eval` when called from `evaluate_model` seems even slower (3-4x slower)? Part of this seems to be because the `all_eval` script sets `padding=False` in `EvalQADataset`, but `evaluate_model` leave it as the default of True. But it's still 2x slower after that.
  - This was comparing an `ascent` forgetting model with a `retain` model - the `ascent` model is slower than the `retain` model for some reason.

src/arcsf/data/data_module.py
jack89roberts: @J-Dymond - I'm still unsure about this, but in two ways:

- Looking at the docs of [torch.Generator.seed](https://pytorch.org/docs/stable/generated/torch.Generator.html#torch.Generator.seed) this will not produce a reproducible value (i.e. the fact the generator is initialised with a seed of 42 has no bearing on what `rand_gen.seed()` will return). I think the correct thing to do would be to use the [`generator` arg of `shuffle`](https://huggingface.co/docs/datasets/v2.20.0/en/package_reference/main_classes#datasets.Dataset.shuffle), but note this is only compatible with `numpy` generators not `torch` generators.

- I'm not sure we want the same perturbed answers to be returned every time this is called with the same `author_n` and `question_n` (within a script), but I'm not sure this is a big issue either.

Some of this may not matter once we have a dataset with actual perturbed answers.

-

src/arcsf/eval/evaluate_model.py
J-Dymond: The base model truth ratios are only used in the forget quality, so we only need them from the forget set. We could load from the retain set as well for completeness, but I don't think they'll be used our the current implementation.

src/arcsf/data/data_module.py
J-Dymond: Ok, I misunderstood this, sorry! I've added another attribute storing the random_seed. But I agree, we don't need this with the new dataset, and I'm not sure we want it either.

J-Dymond (commented):
This was a test, please ignore

jack89roberts (approved):

PR #34, opened 2024-07-26
Add model cache_dir config to Baskerville scripts and related config

### LINKED ISSUE #27: Infrastructure for training large models
e.g. with reference to deep speed code in TOFU codebase

What type of training? (full fine-tuning vs. PEFT etc.)

msmoore:
Added model cacheing support by:

- Including a new <k,v> pair in the 'top config' file;
- Updating the training script generator to include this; and
- Updating the `load_model_and_tokenizer` function to pass it in as part of the model_kwargs expected by the transformers `AutoModelForCausalLM.from_pretrained` method.

### DIFF
--- a/configs/experiment/tofu_test.yaml
+++ b/configs/experiment/tofu_test.yaml
@@ -27,7 +27,9 @@ full_data_config: tofu_full
 
 # Baskerville kwargs
 use_bask: true
-model_cache_dir: /bask/projects/v/vjgo8416-sltv-forget
+model_cache_dir: /bask/projects/v/vjgo8416-sltv-forget/caches/models
+data_cache_dir: /bask/projects/v/vjgo8416-sltv-forget/caches/datasets
+wandb_cache_dir: /bask/projects/v/vjgo8416-sltv-forget/caches/wandb
 bask:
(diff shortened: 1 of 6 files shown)


jack89roberts:
@msmoore Looks good to me (pending giving it a try)! These are secondary as they're much smaller files but would be good to do these too: 
- adding `cache_dir` as an arg (if present in `model_kwargs`) to the `AutoTokenizer.from_pretrained` call in `arcsf.models.model.load_model_and_tokenizer`
- Setting `wandb` cache directories - [these are the two env variables that need to be set](https://github.com/alan-turing-institute/ARC-LoCoMoSeT/blob/c0598a4e0aa1d1e4d7326ff3492ff6831b1a63eb/src/locomoset/config/config_classes.py#L107-L109) (link to another project, see lines 108-109), and can be done in `arcsf.config.experiment.ExperimentConfig.init_wandb` anywhere before the `wandb.init` call.

jack89roberts:
I realised everything can be set with env variables which I'm hoping will avoid needing to make any other tweaks to the code.

jack89roberts (approved):

